PR #3, opened 2020-08-11
store db reader/writer dependencies in starlette app.state

vincentsarago:
This PR moves some variables to the application state (https://www.starlette.io/requests/#other-state). 

cc @geospatial-jeff

### DIFF
--- a/stac_api/utils/dependencies.py
+++ b/stac_api/utils/dependencies.py
@@ -1,7 +1,6 @@
 from dataclasses import dataclass
 from typing import Callable, List
 
-from sqlalchemy.orm import Session
 from starlette.requests import Request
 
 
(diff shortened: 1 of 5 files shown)


stac_api/utils/dependencies.py
```diff
@@ -32,27 +25,27 @@ def _parse(request: Request):
     return _parse
 
 
-def database_reader_factory() -> Session:
+def database_reader_factory(request: Request) -> Session:
     """Instantiate the database reader session"""
     try:
-        if not DB_READER:
+        if not request.app.state.DB_READER:
```
vincentsarago, line 31: I think we can remove this!

geospatial-jeff: Yes!  Same for the writer factory 👍

stac_api/utils/dependencies.py
```diff
@@ -32,27 +25,19 @@ def _parse(request: Request):
...
     """Instantiate the database reader session"""
     try:
-        if not DB_READER:
-            raise DatabaseConnectionError(
-                message="Database engine has not been created"
-            )
-        db = DB_READER()
+        db = request.app.state.DB_READER()
         yield db
     finally:
```
geospatial-jeff, line 33: Can probably remove the `try...finally` as well and close the db connection during the shutdown event.

geospatial-jeff: Which should also let us remove the reader/writer dependencies all together by updating the backend clients to pull the database session directly from the request:
- https://github.com/vincentsarago/arturo-stac-api/blob/storeDepsInState/stac_api/clients/item_crud.py#L134-L135
- https://github.com/vincentsarago/arturo-stac-api/blob/storeDepsInState/stac_api/clients/collection_crud.py#L66-L67
- https://github.com/vincentsarago/arturo-stac-api/blob/storeDepsInState/stac_api/clients/tokens.py#L40-L41

PR #65, opened 2021-01-11
Update broken urls to STAC API spec

kylebarron:
I'm guessing that the STAC API repo was reorganized recently, maybe for 1.0.0-beta.1.


kylebarron:
Might need to re-run CI... Not sure that's related to this change?

lossyrob:
This is based on older changes - does this need to be updated or can this be closed?

geospatial-jeff:
This can be closed

Issue #71, opened 2021-01-19
refactor PostgresClient


geospatial-jeff:

geospatial-jeff:
For context, the initial postgresql application logic (`PostgresCoreClient` et al) was written ~1 year ago in a different repo and refactored very slightly to accommodate the api layer which was added when this project was open sourced (#1).  The original application used a FastAPI dependency to inject the required resources and code into the application.

There are some problems with the current implementation which are starting to block development on other things (notably #57 #58 #64) - in no particular order:
1. [`PostgresClient`](https://github.com/arturo-ai/arturo-stac-api/blob/master/stac_api/clients/postgres/base.py#L18-L70) is overloaded.  It defines methods for database operations like committing, checking if a row exists etc. but these are very opinionated and don't always work for our various use cases.  Overloading of functionality in this particular class [means we use it everywhere](https://github.com/arturo-ai/arturo-stac-api/blob/master/stac_api/clients/tiles/ogc.py#L16-L41), even when it doesn't make much sense to do so.  It's also just very abstracted, making development hard.
2. We are using `dataclass` when we really need to be using `attrs`.  Mostly because attrs is more flexible when it comes to the definition of optional and optionally required attributes.
3. [Defaulting to the default sqlalchemy models](https://github.com/arturo-ai/arturo-stac-api/blob/master/stac_api/clients/postgres/core.py#L37-L38) is weird and causes some not so intuitive behavior, especially when writing certain test cases (#70).  It is also poor separation of concerns.
4. I'm really on the fence about allowing sqlalchemy models as instance attributes of a class.  I'm not sure its bad, but I think there is a better way of doing it.
5. The separation of pagination logic from the core is also very abstracted, and should be brought into core anyways to align with the spec (#67).
6. Coupling of resource management and application logic makes #57 a nightmare.

What I'm going for with the refactor:
1. Separation of resource management from application logic.  [This blog post](https://camillovisini.com/article/abstracting-fastapi-services/) has an interesting implementation using a mixin, not sure if we will take the same approach but its good food for thought.
2. Focus less on DRY.  I'd rather have each database operation manage its own commits and deduplicate code as needed in the future.  This should make the codebase much simpler.
3. Use the newest version of sqlalchemy with support for async (#64)
4. Sqlalchemy models, database configuration etc. are owned by the api layer and passed down to the application logic.

Issue #57, opened 2020-11-21
decouple backends from api layer


geospatial-jeff:
There are still some places where the backend is coupled to the API.  For example, the sqlalchemy engine and session are created during app startup.  This coupling makes it difficult to support additional backends, and forces us to do some hacky things in the code.

A similar treatment was applied to api extensions in https://github.com/arturo-ai/arturo-stac-api/pull/54.

PR #74, opened 2021-01-21
decouple database connection

geospatial-jeff:
Moving the management of database resources to the client, instead of the API.  Ref #57 #71 
- Adds the `stac_api.clients.postgres.session.Session` object which uses `fastapi-utils` to manage database sessions and commits.
    - Previously, database sessions were managed by a FastAPI middleware and some context vars to distribute the session throughout the app.  Commits were handled through `PostgresClient` which has been removed.
    - `PaginationTokenClient` is now a mixin to align with the spec (pagination is now part of core) and to re-use the database session of the parent class.
    - Note how the app factory (`stac_api.api.app`) no longer references any postgres related code.


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.8-slim
 
 # Any python libraries that require system libraries to be installed will likely
 # need the following packages in order to build
-RUN apt-get update && apt-get install -y build-essential git
+RUN apt-get update && apt-get install -y build-essential git libgeos-dev
```
vincentsarago, line 5: wheels are now available for pygeos 0.9.0

stac_api/clients/postgres/session.py
```diff
@@ -0,0 +1,53 @@
...
+        """override base method to include exception handling"""
+        try:
+            yield from self.get_db()
+        except sa.exc.IntegrityError as e:
+            if isinstance(e.orig, psycopg2.errors.UniqueViolation):
+                raise errors.ConflictError("resource already exists") from e
+            elif isinstance(e.orig, psycopg2.errors.ForeignKeyViolation):
+                raise errors.ForeignKeyError("collection does not exist") from e
+            logger.error(e, exc_info=True)
+            raise errors.DatabaseError("unhandled database error")
```
anlaksh88, line 36: Do we need a catch-all exception handler here?
```suggestion
            raise errors.DatabaseError("unhandled database error")
       except Exception as e:
            logger.error(e, exc_info=True)
            raise errors.DatabaseError("unhandled database error")
         
```

geospatial-jeff: yes good catch, we need to handle exceptions from `get_db` which are not of `IntegrityError`.

geospatial-jeff: catch-all exception handler doesn't work here because there are other exceptions that might get thrown within the context of a database session (like `NotFoundError`).  These exceptions end up getting lost in the catch-all exception handler which then impacts response codes.

Instead I've updated the `except` to use a lower-level sqlalchemy exception.  It's the lowest level that includes the wrapped psycopg2 error (`e.orig`) that we need to determine what error to throw.  It should cover [any error which occurs during the execution of a SQL statement](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/exc.py#L377-L387), which I think covers our requirements here.

anlaksh88 (approved):

PR #75, opened 2021-01-26
fix precommit

geospatial-jeff:
- Fixes precommit config so black/isort don't conflict.
- Re-lints files.


.pre-commit-config.yaml
```diff
@@ -22,7 +21,7 @@ repos:
                   # E501 let black handle all line length decisions
                   # W503 black conflicts with "line break before operator" rule
                   # E203 black conflicts with "whitespace before ':'" rule
-                  '--ignore=E501,W503,E203']
+                  '--ignore=E501,W503,E203,C901']
```
kylebarron, line 24: Just use `profile=black` for isort. https://pycqa.github.io/isort/docs/configuration/black_compatibility/

kylebarron: Oh nevermind I see that elsewhere in the config.

kylebarron (approved):

PR #82, opened 2021-01-31
add docstrings

geospatial-jeff:
Adding google style docstrings to API layer + base classes and patterns.  Modifying other docstrings enough to make `pydocstyle` happy.

### DIFF
--- a/stac_api/api/app.py
+++ b/stac_api/api/app.py
@@ -180,7 +180,6 @@ class StacApi:
 
     def customize_openapi(self) -> Optional[Dict[str, Any]]:
         """Customize openapi schema."""
-        # """customize openapi schema"""
         if self.app.openapi_schema:
             return self.app.openapi_schema
 
--- a/stac_api/api/app.py
+++ b/stac_api/api/app.py
@@ -1,5 +1,5 @@
 """fastapi app creation"""
-from typing import Dict, List, Optional, Type
+from typing import Any, Dict, List, Optional, Type
 
 import attr
 from fastapi import APIRouter, FastAPI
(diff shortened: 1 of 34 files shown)


kylebarron (approved):
- Might not be a bad idea to include a docstring linter so that regressions are caught, e.g. `pydocstring`
- Nit: I usually strive for the first line of the docstring (e.g. right after the `"""` to be within 80 characters, so I'd personally put URLs after the first line of the docstring.
- I don't think it's necessarily important to include the types _twice_. I think if you're typing function arguments, it's probably ok to leave it out from the docstring string itself. I.e. my editor tells input types 
image
- If you do include the types in the docstring, I think it's fine to use Python typing syntax, e.g. `Set[str]`, not just `set`

stac_api/api/app.py
```diff
@@ -136,8 +178,9 @@ def register_core(self):
         )
         self.app.include_router(router)
 
-    def customize_openapi(self):
-        """customize openapi schema"""
+    def customize_openapi(self) -> Optional[Dict[str, Any]]:
+        """Customize openapi schema."""
+        # """customize openapi schema"""
```
vincentsarago, line 183: 👀

vincentsarago, line 183:
```suggestion
```

stac_api/api/app.py
```diff
@@ -150,18 +193,24 @@ def customize_openapi(self):
...
         self.app.include_router(mgmt_router, tags=["Liveliness/Readiness"])
 
     def __attrs_post_init__(self):
-        """post-init hook"""
+        """Post-init hook.
+
+        Responsible for setting up the application upon instantiation of the class.
+
+        Returns:
+            None
```
vincentsarago, line 212: is the return statement mandatory?

stac_api/clients/base.py
```diff
@@ -15,86 +15,183 @@
...
+    """BulkTransactionsClient."""
 
     @staticmethod
     def _chunks(lst, n):
-        """Yield successive n-sized chunks from lst."""
+        """Yield successive n-sized chunks from list.
+
+        https://stackoverflow.com/questions/312443/how-do-you-split-a-list-into-evenly-sized-chunks
+        """
+        # """Yield successive n-sized chunks from lst."""
```
vincentsarago, line 124:
```suggestion
```

vincentsarago (approved):

kylebarron (approved):

PR #95, opened 2021-02-18
Create .dockerignore

### LINKED ISSUE #91: add a .dockerignore

geospatial-jeff:
closes #91


kylebarron (approved):
Just a note that I've previously been bitten by dockerignore not matching nested folders without `**`, like I think as is `stac_api/__pycache__` wouldn't be ignored unless changed to `**/__pycache__`. Ref https://stackoverflow.com/questions/43747776/copy-with-docker-but-with-exclusion#comment89807993_43747867

PR #97, opened 2021-02-24
WIP connect titiler for custom search tiling

kylebarron:
This is a draft PR to connect titiler with STAC api's search endpoint.

Notes:

- This only cares about the `tile` endpoint of the `MosaicTilerFactory`; I haven't tested anything else.
- This adds an extension file, but doesn't add it to the main app by default because that would require adding more dependencies: `cogeo_mosaic`, `morecantile`, `rio_tiler`. (Though I guess the STAC API mounts `titiler` by default already, so I guess this isn't actually adding any more dependencies?)

### DIFF
--- a/stac_api/api/extensions/tiles_search.py
+++ b/stac_api/api/extensions/tiles_search.py
@@ -1,0 +1,231 @@
+"""transaction extension."""
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Tuple, Type
+
+import attr
+from cogeo_mosaic.backends.base import BaseBackend
+from cogeo_mosaic.backends.stac import default_stac_accessor
+from cogeo_mosaic.errors import NoAssetFoundError
+from cogeo_mosaic.mosaic import MosaicJSON
+from fastapi import FastAPI, Query
+from morecantile import TileMatrixSet
+from rio_tiler.constants import WEB_MERCATOR_TMS
+from rio_tiler.io import BaseReader, STACReader
+from rio_tiler.models import ImageData
+from rio_tiler.mosaic import mosaic_reader
+from titiler.dependencies import DefaultDependency
+from titiler.endpoints.factory import MosaicTilerFactory
+
+from stac_api.api.extensions.extension import ApiExtension
+from stac_api.clients.postgres.core import CoreCrudClient
+
+
+@attr.s
+class FakeRequest:
+    base_url: str = "http://localhost:8081"
+    query_params: Dict = {}
+
+
+@attr.s
+class DynamicStacBackend(BaseBackend):
+    """Like a STAC backend but dynamic"""
+
+    path: str = attr.ib()
+    reader: Type[BaseReader] = attr.ib(default=STACReader)
+    reader_options: Dict = attr.ib(factory=dict)
+    backend_options: Dict = attr.ib(factory=dict)
+
+    item_collection: Dict = attr.ib(factory=dict)
+
+    # default values for bounds and zoom
+    bounds: Tuple[float, float, float, float] = attr.ib(default=(-180, -90, 180, 90))
+    minzoom: int = attr.ib(default=0)
+    maxzoom: int = attr.ib(default=30)
+    mosaic_quadkey_zoom: Optional[int] = attr.ib(default=None)
+
+    # Because we are not using mosaicjson we are not limited to the WebMercator TMS
+    tms: TileMatrixSet = attr.ib(default=WEB_MERCATOR_TMS)
+
+    mosaic_def: MosaicJSON = attr.ib(init=False)
+
+    _backend_name = "DynamicSTAC"
+
+    def __attrs_post_init__(self):
+        """Post Init."""
+        # Construct a FAKE mosaicJSON
+        self.mosaic_def = MosaicJSON(
+            mosaicjson="0.0.2",
+            name="it's fake but it's ok",
+            minzoom=self.minzoom,
+            maxzoom=self.maxzoom,
+            quadkey_zoom=self.mosaic_quadkey_zoom,
+            tiles=[],
+        )
+
+    def write(self, overwrite: bool = True):
+        """Write mosaicjson document."""
+        pass
+
+    def update(self):
+        pass
+
+    def _read(self) -> MosaicJSON:
+        pass
+
+    def assets_for_tile(self, x: int, y: int, z: int, **kwargs) -> List[str]:
+        """Retrieve assets for tile."""
+        bounds = self.tms.bounds(x, y, z)
+        return self.get_assets(list(bounds), **kwargs)
+
+    def assets_for_point(self, lng: float, lat: float) -> List[str]:
+        """Retrieve assets for point."""
+        # geom = {"type": "Point", "coordinates": [lng, lat]}
+        bounds = [lng, lat, lng, lat]
+        return self.get_assets(bounds)
+
+    def get_assets(self, bbox, **kwargs) -> List[str]:
+        """Find assets."""
+        print("get assets kwargs")
+        print(kwargs)
+        kwargs["bbox"] = bbox
+        client = self.reader_options["client"]
+
+        kwargs["request"] = FakeRequest()
+
+        print("get_assets edited kwargs")
+        print(kwargs)
+
+        feature_collection = client.get_search(**kwargs)
+
+        print("feature_collection")
+        print(feature_collection)
+        return [default_stac_accessor(f) for f in feature_collection["features"]]
+
+    @property
+    def _quadkeys(self) -> List[str]:
+        return []
+
+    def tile(  # type: ignore
+        self,
+        x: int,
+        y: int,
+        z: int,
+        reverse: bool = False,
+        **kwargs: Any,
+    ) -> Tuple[ImageData, List[str]]:
+        """Get Tile from multiple observation."""
+        print("tile kwargs")
+        print(kwargs)
+        mosaic_assets = self.assets_for_tile(x, y, z, **kwargs)
+        if not mosaic_assets:
+            raise NoAssetFoundError(f"No assets found for tile {z}-{x}-{y}")
+
+        if reverse:
+            mosaic_assets = list(reversed(mosaic_assets))
+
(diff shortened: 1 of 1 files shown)


stac_api/api/extensions/tiles_search.py
```diff
@@ -0,0 +1,231 @@
...
+        None, title="Collections", description="Collections to search within"
+    )
+
+    ids: Optional[str] = Query(None)
+    datetime: Optional[str] = Query(None)
+    limit: Optional[int] = Query(10)
+    query: Optional[str] = Query(None)
+    token: Optional[str] = Query(None)
+    fields: Optional[str] = Query(None)
+    sortby: Optional[str] = Query(None)
```
kylebarron, line 179: todo: add title and description for each of these, so it shows up in openapi

stac_api/api/extensions/tiles_search.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+@attr.s
+class TilesSearchExtension(ApiExtension):
+    """Custom Titiler Extension.
+
+    Custom extension for sending query results to a titiler DynamicSTAC endpoint
+    """
+
+    client: CoreCrudClient = attr.ib(default=attr.Factory(CoreCrudClient))
```
kylebarron, line 208: This uses the base `CoreCrudClient` as all it needs is the `client.get_search` method. I figured it wasn't necessary to subclass `CoreCrudClient` like the `TilesClient` does

https://github.com/arturo-ai/arturo-stac-api/blob/7ca0c8790ac64bbd359e88933ef94f25ece5b383/stac_api/clients/tiles/ogc.py#L17-L41

stac_api/api/extensions/tiles_search.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Args:
+            app: target FastAPI application.
+
+        Returns:
+            None
+        """
+        dynamic_stac_endpoint = MosaicTilerFactory(
+            reader=DynamicStacBackend,
+            reader_options={"client": self.client},
```
kylebarron, line 221: What's the best way to pass the `client` to the `MosaicTilerFactory`? I need to access the client in the `MosaicBackend.get_assets` method.

stac_api/api/extensions/tiles_search.py
```diff
@@ -0,0 +1,231 @@
...
+        kwargs["request"] = FakeRequest()
+
+        print("get_assets edited kwargs")
+        print(kwargs)
+
+        feature_collection = client.get_search(**kwargs)
+
+        print("feature_collection")
+        print(feature_collection)
+        return [default_stac_accessor(f) for f in feature_collection["features"]]
```
kylebarron, line 102: I don't fully understand this... seems like here I'm returning the _id_ of each stac item, rather than the stac items themselves... Seems like it would be better to return `feature_collection['features']` and then use the resulting STAC dicts directly.

stac_api/api/extensions/tiles_search.py
```diff
@@ -0,0 +1,231 @@
...
+from rio_tiler.mosaic import mosaic_reader
+from titiler.dependencies import DefaultDependency
+from titiler.endpoints.factory import MosaicTilerFactory
+
+from stac_api.api.extensions.extension import ApiExtension
+from stac_api.clients.postgres.core import CoreCrudClient
+
+
+@attr.s
+class FakeRequest:
```
kylebarron, line 24: This is to mock the `request` param passed through by default here https://github.com/arturo-ai/arturo-stac-api/blob/7ca0c8790ac64bbd359e88933ef94f25ece5b383/stac_api/api/routes.py#L58-L66

stac_api/api/extensions/tiles_search.py
vincentsarago: > but it seems like it shouldn't be on the reader_options.

I think it's ok to use the reader_options here and just pass a `backend_options` dict within. 

```python
reader_options={"backend_options":{"client": self.client}},
```

that's how we do for STAC https://github.com/developmentseed/cogeo-mosaic/blob/master/cogeo_mosaic/backends/stac.py#L76

I agree it's not perfect

vincentsarago: Wait, in fact you are already creating a custom Mosaic Backend so you can still use `reader_options={"client": self.client}` and make client a required parameter for your `DynamicStacBackend` backend

stac_api/api/extensions/tiles_search.py
```diff
@@ -0,0 +1,235 @@
...
+            reader_options={"client": self.client},
+            # path_dependency=MosaicParams,
+            dataset_reader=STACReader,
+            dataset_dependency=DatasetDependency,
+            layer_dependency=AssetParams,
+            router_prefix="test-dynamic-search",
+        )
+
+        app.include_router(
+            dynamic_stac_endpoint.router, prefix="/dynamic-stac", tags=["Dynamic STAC"]
```
kylebarron, line 234: It seems unnecessary to have both of these router prefixes, and it looks like `/dynamic-stac` overrides `test-dynamic-search` anyways. Also, open to other prefix names.

geospatial-jeff:
To summarize our slack convo:

The "all-in-one" approach where we overload a single endpoint to `POST /search` and return tiles is simpler to implement but doesn't align that well with the way that others typically implement STAC + tiling.  The typical approach is to add a new link which references a tile layer for that particular resource.

```json
{
  "title": "tile-layer",
  "type": "image/jpeg",
  "href": "http://tile-layer/{z}/{x}/{y}.jpg",
  "templated": true,
  "rel": "tiles",
}
```

In my mind this has two primary benefits:
- Its more canonical "STAC", where additional info/metadata/functionality etc. is exposed as a link on the particular resource.
- It lets us hook into specifications like OGC API Tiles which has a very sensible and small core and provides nice interoperable way of linking STAC to tiles.  This is actually how the existing titiler extensions works, albeit only on a single item so its a much easier problem.

My dream for STAC + tiling is exposing `tiles` links at the Item, ItemCollection, and Catalog levels; where item collection and catalog links are mosaicing the appropriate items.  The actual implementation of this is difficult though.  And the "all-in-one" approach provides a much more constrained example that I think we should move forward with because it should give us valuable information about how to implement STAC + tiling as described above.

In particular I think breaking up this repo into appropriate parts (stac-api / db / extension - ref @vincentsarago) will clear up separation of concerns and make thinking through the implementation a little easier.

stac_api/api/extensions/tiles_search.py
geospatial-jeff: A way to get around this is to inject a dependency which itself returns a request object object.  This works because dependencies always have access to the request:

```python

def request_dependency(request: Request):
    return request
```

stac_api/api/extensions/tiles_search.py
geospatial-jeff: makes sense.  `TilesClient` arguably shouldn't subclass `CoreCrudClient` either.  I think composition is a better pattern here.

stac_api/api/extensions/tiles_search.py
geospatial-jeff: it is unnecessary, I think i'd rather include the prefix on the call to `include_router`.

kylebarron:
In general I think it makes sense to put `tiling` rel links on `Item`, `ItemCollection`, and `Catalog`, however it seems you hit unnecessary back-and-forth HTTP requests if the client needs to first create an `ItemCollection` via the `/search` endpoint, then post those stac items back to tile them.

> exposing `tiles` links at the Item, ItemCollection, and Catalog levels; where item collection and catalog links are mosaicing the appropriate items

I think this makes sense, but in my mind this PR is essentially intending to serve the "tile a Catalog" problem. For almost any Catalog, like Landsat 8/Sentinel 2, you have way too many images per relevant tile, so you still need a search within the catalog.

kylebarron:
> The typical approach is to add a new link which references a tile layer for that particular resource.
> 
> ```json
> {
>   "title": "tile-layer",
>   "type": "image/jpeg",
>   "href": "http://tile-layer/{z}/{x}/{y}.jpg",
>   "templated": true,
>   "rel": "tiles",
> }
> ```

By the way this isn't necessarily in conflict with a `/search/tiles` approach. A `/search` endpoint that returns an `ItemCollection` could have a `tiles` link back to `/search/tiles` in order to tile the contained features. And it could set `/search` params like `limit` and `token` so that those exact same features would be used. In effect it would mostly trade another `SELECT` against the db for avoiding `POST`ing the `ItemCollection` back to a tiler.

geospatial-jeff:
> I think this makes sense, but in my mind this PR is essentially intending to serve the "tile a Catalog" problem.

Makes sense to me.  Catalog tiling should happen through `/search?collections=collection_name` 

> By the way this isn't necessarily in conflict with a `/search/tiles` approach. A `/search` endpoint that returns an `ItemCollection` could have a `tiles` link back to `/search/tiles` in order to tile the contained features.

This is a great approach to make this more consumable, let's roll with it.

vincentsarago:
FYI, I've started something similar over https://github.com/stac-utils/titiler-pgstac 

Note: This is not an extension and doesn't add new endpoint to a stac-api but create a new standalone app

lossyrob:
What's the status of this PR? IMO stac-fastapi shouldn't host its own tiler - the titiler-pgstac is a good example of standing up a tiler that uses STAC but a separate service that can be deployed independently. In our implementation of stac-fastapi, we override the pgstac backend core client in order to iterate over collections/items and inject tiler links where appropriate. I think this approach allows stac-fastapi to concentrate on just the STAC API endpoints, but allows customization by users if they want to add titiler or other tiling capability. I could see work done to generalize and implement a link injection mechanism so that other libraries can write hooks into the STAC objects returned by stac-fastapi so that the link/asset injection logic can live in that external project. But I don't think we should create tilers directly in stac-fastapi.

geospatial-jeff:
I agree the code for such an integration should live somewhere else.  Closing this PR.

> I could see work done to generalize and implement a link injection mechanism so that other libraries can write hooks into the STAC objects returned by stac-fastapi so that the link/asset injection logic can live in that external project.

This would be a good feature.

PR #101, opened 2021-03-02
Docker Changes

bitner:
- Add extra_hosts to docker-compose so that host.docker.internal works on Linux systems
- Change POSTGRES_PORT to 5439 so that it does not conflict with default Postgres installs on host systems


kylebarron:
Should close https://github.com/arturo-ai/arturo-stac-api/issues/63

geospatial-jeff (approved):
LGTM!

Issue #69, opened 2021-01-12
use a single endpoint factory


geospatial-jeff:
https://github.com/arturo-ai/arturo-stac-api/blob/fb47dedfbc45df4488f7fa169b76ca1b30a420f1/stac_api/api/routes.py

Endpoint factories wrap a callable in a function that can be executed as a FastAPI route.  Doing so lets us "decorate" the callable with specific request/response models.  Currently we have two factories, one for routes which define the request using a `dataclass` and one as a pydantic model.
- Dataclasses are used because of their support for dependency injection.
- Pydantic models are used for static types (no dependency injection).

It would be much better to have a single factory instead which means either (1) use a single request type for all routes or (2) one factory that can understand both dataclass + pydantic models.

geospatial-jeff:
Closed with https://github.com/stac-utils/stac-fastapi/pull/126

PR #119, opened 2021-03-30
support async endpoints

### LINKED ISSUE #116: specify async endpoints
For asyncpg or sqlalchemy>1.4 we need to be able to specify `async def` endpoints so the code may be executed by the event loop rather than a background thread.  Right now it only supports `def` (sync) endpoints.

https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi_api/stac_fastapi/api/routes.py#L29

geospatial-jeff:
Adds support for async endpoints, required if the client code is async (ex. asyncpg).

This adds the `endpoint_factory` instance attribute everywhere we are creating routes (the main API, transactions/tiles/bulk_transactions extensions).  It allows the caller to specify their own endpoint factory.  Also added is the `create_async_endpoint` function uses `async def` route handlers.  The attribute defaults to using sync route handlers, but this may change if the sqlalchemy implementation is upgraded to 1.4 which does support async.

```python
from stac_fastapi.api.routes import create_async_endpoint

api = StacApi(
    ...,
    endpoint_factory=create_async_endpoint
)
app = api.app
```

cc @bitner 

Closes #116 #69

### DIFF
--- a/stac_fastapi_api/stac_fastapi/api/app.py
+++ b/stac_fastapi_api/stac_fastapi/api/app.py
@@ -16,10 +16,7 @@ from stac_fastapi.api.models import (
     SearchGetRequest,
     _create_request_model,
 )
-from stac_fastapi.api.routes import (
-    create_endpoint_from_model,
-    create_endpoint_with_depends,
-)
+from stac_fastapi.api.routes import create_endpoint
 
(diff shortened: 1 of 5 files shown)


bitner:
Perfect! Was just running into this! I'll test out and report.

PR #125, opened 2021-04-15
Convert to using namespace packaging and setup.py

bitner:
So, I'm not sure what the full intention was behind making each submodule it's own package, but needing to recursively go in and install things (in the right order) was driving me bonkers. I also have to admit little/no familiarity with using pipenv/Pipfile for managing dependencies.

This PR moves all the stac_fastapi_* directories to where they were simlinked under stac_fastapi.

I *think* this should still be able to meet the intent of the initial move to submodules and be a bit easier. For example the way things are set up with this PR `pip install -e .[dev,sqlalchemy,tiles]` will install stac_fastapi along with the sqlalchemy and tiles dependencies. When I make a PR with the pgstac work, if you wanted to use pgstac rather than sqlalchemy, running `pip install -e .[dev,pgstac,tiles]` would still end up having the stac_fastapi.sqlalchemy code, but it would at least avoid the install of psycopg2/sqlalchemy etc that are only used with stac_fastapi.sqlalchemy.

Happy to chat through this one, but at least to me, this seems way easier, but, then again, I might be missing something entirely with the original intent.


bitner:
@geospatial-jeff ^^

bitner:
arrrg, there's something up with black -- pre-commit passing on my end, but there's a line-length discrepency with what is running on the checks

bitner:
PR Refactored to create separate namespace packages.

Tests right now have all been moved to sqlalchemy package to get to run. Tests that do not rely on the sqlalchemy should be moved into the appropriate packages.

Dockerfile
```diff
@@ -1,55 +1,34 @@
...
-RUN pipenv install --deploy --ignore-pipfile ${install_dev_dependencies:+--dev}
+COPY . /app
+
+ENV PATH=$PATH:/install/bin
 
+RUN mkdir -p /install && \
+    pip install -e ./stac_fastapi/api[dev]   && \
+    pip install -e ./stac_fastapi/types[dev]  && \
+    pip install -e ./stac_fastapi/extensions[dev,tiles]  && \
+    pip install -e ./stac_fastapi/sqlalchemy[dev,server]
```
kylebarron, line 23: Why use `-e` in a docker setup? Will docker see local changes?

stac_fastapi/extensions/setup.py
```diff
@@ -0,0 +1,56 @@
...
+    os.path.join(os.path.dirname(__file__), "stac_fastapi/extensions/version.py"),
+).__version__  # type:ignore
+
+install_requires = [
+    "fastapi",
+    "attrs",
+    "pydantic[dotenv]",
+    "stac_pydantic==1.3.8",
+    "stac-fastapi.types",
+    "stac-fastapi.api",
```
kylebarron, line 22: In the future we might want to do

```
f"stac-fastapi.types=={__version__}"
```

if we think it's important users have the same versions installed.

stac_fastapi/pgrest/setup.py
```diff
@@ -0,0 +1,65 @@
...
+        "fastapi-utils",
+    ],
+    "dev": ["pytest", "pytest-cov", "pytest-asyncio", "pre-commit", "requests"],
+    "tiles": ["titiler==0.2.*"],
+    "docs": ["mkdocs", "mkdocs-material", "pdocs"],
+}
+
+
+setup(
+    name="stac-fastapi",
```
kylebarron, line 42:
```suggestion
    name="stac-fastapi.pgrest",
```
?

stac_fastapi/types/setup.py
```diff
@@ -0,0 +1,53 @@
...
+from setuptools import find_namespace_packages, setup
+
+with open("README.md") as f:
+    desc = f.read()
+
+# Get version from stac-fastapi-api
+__version__ = load_source(
+    "stac_fastapi.types.version",
+    os.path.join(os.path.dirname(__file__), "stac_fastapi/types/version.py"),
+).__version__  # type:ignore
```
kylebarron, line 14: Might be simpler to have a `version.txt` file at the repo root, or maybe consider something like [`versioneer`](https://github.com/python-versioneer/python-versioneer)

tox.ini
```diff
@@ -11,7 +11,6 @@ max-line-length = 90
 
 [tool:isort]
 profile=black
-known_arturo=arturo
```
kylebarron, line 14: 👀

kylebarron (approved):

kylebarron:
So this is moving

- `stac_fastapi_api`
- `stac_fastapi_extensions`
- `stac_fastapi_server`, etc

into `stac_fastapi/{api,extensions,server}`?

Dockerfile
bitner: for me that's the biggest point of having the docker setup! docker-compose is set to mount the local directory so I can use it for development along with a database I can wipe at will.

stac_fastapi/pgrest/setup.py
bitner: side effect of doing some bulk operations to bootstrap things - pgstac does not even exist in this branch yet

Dockerfile
kylebarron: That's really interesting. I've only ever used Docker for deployment, not development.

bitner:
@kylebarron I have no strong opinions on the folder layout - as discussed on slack, I was directly copying the layout from titiler.

kylebarron:
I don't have any strong opinions either, was just trying to make clear that there are flexible naming options.

vincentsarago:
In Titiler we added a `top level` metapackage (no code, just a setup.py) to let the user still do `pip install titiler` https://github.com/developmentseed/titiler/issues/294.

Also noting that https://github.com/bitner/stac-fastapi/blob/onesetup/scripts/publish  will need to get updated with the correct package names

geospatial-jeff:
The `server` submodule was removed (I think accidentally), could we add that back?  Also I think we might need a `setup.py` at the top level of the repo (at least titiler has one).  Otherwise LGTM!

geospatial-jeff:
> a server only makes sense in the context of having a backend

Oh I see, the serving code was just moved inside the backend package not removed entirely.  I think your argument is valid, especially considering the serving code will be different across backends so it doesn't make sense to have the serving code as its own package.

> to have no top level metapackage

I cocanld see arguments for both.  Its certainly convenient to be able to install all the namespace packages through a top-level metapackage, but I'm not sure how practical it is, especially as the project supports multiple backends.  I think we leave out the metapackage for now, and add one in the future if it becomes needed.

bitner:
RE: no top level metapackage

To install everything, means having two apps and two backends which are mutually exclusive which I think is more confusing than convenient. If anything, we could pick which backend was a default and have the metapackage install that, but I think that could get confusing when someone decided they wanted to other backend instead. I just prefer making people be explicit.

geospatial-jeff (approved):

PR #129, opened 2021-04-28
set env variable in doc actions

vincentsarago:
Without the Env set we get model validation error

### DIFF
--- a/.github/workflows/deploy_mkdocs.yml
+++ b/.github/workflows/deploy_mkdocs.yml
@@ -47,6 +47,8 @@ jobs:
           POSTGRES_DBNAME: postgis
           POSTGRES_HOST: localhost
           POSTGRES_PORT: 5432
+          POSTGRES_HOST_READER: localhost
+          POSTGRES_HOST_WRITER: localhost
 
       - name: Deploy docs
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_mkdocs.yml
```diff
@@ -40,7 +41,12 @@ jobs:
             --exclude_source \
             --overwrite \
             stac_fastapi
-
+        env:
+          POSTGRES_USER: username
+          POSTGRES_PASS: password
+          POSTGRES_DBNAME: postgis
+          POSTGRES_HOST: localhost
```
geospatial-jeff, line 48: this needs to be `POSTGRES_HOST_READER` and `POSTGRES_HOST_WRITER`

.github/workflows/deploy_mkdocs.yml
```diff
@@ -40,7 +41,12 @@ jobs:
...
+          POSTGRES_PORT: 5432
```
vincentsarago, line 49:
```suggestion
          POSTGRES_HOST_READER: localhost
          POSTGRES_HOST_WRITER: localhost
```

geospatial-jeff (approved):

PR #138, opened 2021-05-27
Include collection ID in ItemURI

### LINKED ISSUE #135: `ItemUri`s should provide collection IDs as well as item IDs
The URL structure for the STAC API makes it clear that collection IDs are potentially required to access items by their ID. A small change to the `ItemUri` model should do the trick: https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/stac_fastapi/api/models.py#L66

moradology:
This PR ensures all URL parameters provided to get a STAC item are available for implementing `get_item` methods. Previously, only the item id was available.

Closes #135

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/models.py
+++ b/stac_fastapi/api/stac_fastapi/api/models.py
@@ -70,7 +70,7 @@ class ItemUri(CollectionUri):
 
     def kwargs(self) -> Dict:
         """kwargs."""
-        return {"id": self.itemId}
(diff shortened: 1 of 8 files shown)


geospatial-jeff (approved):

geospatial-jeff:
This will be a breaking change because we are changing the signature of the base class.  I'm fine with this because it's more explicit and it's better to forward all path parameters included in the request.

PR #139, opened 2021-05-27
Add title configuration to service and core client

### LINKED ISSUE #136: Provide configurable title/version for service
`StacAPI` should accept title and version parameters which should then be used in OpenAPI generation

moradology:
This PR adds configurability for title, version, and description.

These changes are fairly straightforward but, as a result of the fact that `client` is a required (non-default) argument, it is not possible to thread a single title through from service to client. It might be more sensible to put title, version, and description on the settings object to avoid duplication. On the other hand, it might be useful to allow OpenAPI and the LandingPage to have different titles. Thoughts?

Closes #136

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -60,6 +60,8 @@ class StacApi:
         default=attr.Factory(lambda: DEFAULT_STATUS_CODES)
     )
     app: FastAPI = attr.ib(default=attr.Factory(FastAPI))
+    title: str = attr.ib(default="Arturo STAC API")
+    version: str = attr.ib(default="0.1")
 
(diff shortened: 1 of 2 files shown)


stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -190,7 +192,7 @@ def customize_openapi(self) -> Optional[Dict[str, Any]]:
 
         # TODO: parametrize
```
lossyrob, line 193: Remove TODO

lossyrob (changes requested):

stac_fastapi/api/stac_fastapi/api/app.py
moradology: Good catch.

geospatial-jeff:
LGTM!  Let's please hold off on merging this until after tomorrow's discussion on custom models :)

PR #141, opened 2021-05-27
Use "items", not "item" in collection links.

lossyrob:
The link that is currently returned in Collections for the items endpoint is rel="item", but should be rel="items".

Related PySTAC issue: https://github.com/stac-utils/pystac/issues/343


geospatial-jeff (approved):
lgtm!  Do you think we should update the `Relations` enum in stac-pydantic?

PR #143, opened 2021-06-01
Avoid use of leading slash for urljoins

### LINKED ISSUE #142: Ensure urljoins with base_url use relative paths
In several instances a `urljoin` is used with Fast API's `base_url`, where a leading `/` is used during the join. This works if there's no `root_path` set, but in the case where `base_url` contains a path prefix, the leading / makes the resulting join based on the host information and disregards the root_path.

Examples:
[BaseLinks.root](https://github.com/stac-utils/stac-fastapi/blob/178a625996bd08a00e4c5ec6af0ef5731d077e9e/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/models/links.py#L29) - joining to "/" erases the root_path, should just use `str(self.base_url)`
- [CollectionLinks.parent](https://github.com/stac-utils/stac-fastapi/blob/178a625996bd08a00e4c5ec6af0ef5731d077e9e/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/models/links.py#L45)
- [ItemLinks.self](https://github.com/stac-utils/stac-fastapi/blob/178a625996bd08a00e4c5ec6af0ef5731d077e9e/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/models/links.py#L76)  - removing the leading slash will fix this.

The goal of this issue is to find all the instances where a urljoin is used with a leading slash, and joining to a relative path instead (or avoiding a join in the case where the `base_url` can be used directly).

moradology:
This PR removes all leading slashes from uses of `urljoin`, thus allowing users to specify a `root_path` without it being ignored in the process of joining

Closes #142

### DIFF
--- a/scripts/ingest_joplin.py
+++ b/scripts/ingest_joplin.py
@@ -15,7 +15,7 @@ def ingest_joplin_data(data_dir=Path.cwd() / "tests" / "data" / "joplin"):
     with open(data_dir / "collection.json") as f:
         collection = json.load(f)
 
-    r = requests.post(urljoin(app_host, "/collections"), json=collection)
+    r = requests.post(urljoin(app_host, "collections"), json=collection)
     if r.status_code not in (200, 409):
         r.raise_for_status()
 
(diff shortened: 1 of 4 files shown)


geospatial-jeff (approved):
lgtm!  Please lint the code so it passes checks.

geospatial-jeff:
Thanks @moradology !

PR #134, opened 2021-05-24
Prototype model customization

moradology:
This draft PR is not meant for merger but to provide an example of what would be necessary under the current `SqlAlchemy` backend to support model customization. This draft is relevant for further investigating a solution to #58 and #121.

At issue is the fact that some properties in STAC `Item`s and `Collection`s are meant to live at the top level of said `Item`s and `Collection`s - full-fledged STAC extensions but also the grey area of custom models which are technically not fully implemented as STAC extensions. Providing top level support of this kind means subclassing the Pydantic models used by `stac-fastapi` to provide type information for `fastapi`'s automatic documentation generation. Unfortunately, this concession doesn't neatly fit for a few reasons. These limitations are discussed below. First, some example output from the collections endpoint as instantiated by `Docker Compose`:

```json
{
  "id":"joplin",
  "description":"This imagery was acquired by the NOAA Remote Sensing Division to support NOAA national security and emergency response requirements. In addition, it will be used for ongoing research efforts for testing and developing standards for airborne digital imagery. Individual images have been combined into a larger mosaic and tiled for distribution. The approximate ground sample distance (GSD) for each pixel is 35 cm (1.14 feet).",
  "stac_version":"1.0.0-beta.2",
  "links":[
    {
      "href":"http://localhost:8081/collections/joplin",
      "rel":"self",
      "type":"application/json",
      "title":null,
      "label":null
    },
    {
      "href":"http://localhost:8081/",
      "rel":"parent",
      "type":"application/json",
      "title":null,
      "label":null
    },
    {
      "href":"http://localhost:8081/collections/joplin/items",
      "rel":"item",
      "type":"application/geo+json",
      "title":null,
      "label":null
    },
    {
      "href":"http://localhost:8081/",
      "rel":"root",
      "type":"application/json",
      "title":null,
      "label":null
    }
  ],
  "stac_extensions":[
    
  ],
  "title":null,
  "license":"public-domain",
  "extent":{
    "spatial":{
      "bbox":[
        [
          -94.6911621,
          37.0332547,
          -94.402771,
          37.1077651
        ]
      ]
    },
    "temporal":{
      "interval":[
        [
          "2000-02-01T00:00:00Z",
          "2000-02-12T00:00:00Z"
        ]
      ]
    }
  },
  "keywords":null,
  "providers":null,
  "summaries":null,
  "sci_doi":"default_doi",
  "sci_citation":"default_citation",
  "sci_publications":"publication1, publication2, publication3"
}
```

[...]

PR #147, opened 2021-06-13
support custom models

### LINKED ISSUE #58: support custom data models
Currently there is no good way to use a different data model than what is defined in `models/database.py`.  Currently this is difficult to change in a way that is sustainable long term, should be much easier once #57 is resolved.

geospatial-jeff:
closes #58 

See https://github.com/stac-utils/stac-fastapi/pull/147#issuecomment-882111611 for more info.


stac_fastapi/types/stac_fastapi/types/stac.py
```diff
@@ -0,0 +1,56 @@
...
+    type: str
+    stac_version: str
+    stac_extensions: Optional[List[str]]
+    id: str
+    title: Optional[str]
+    description: str
+    links: List[Dict[str, Any]]
+
+
+class Collection(Catalog):
```
geospatial-jeff, line 29: shouldn't use inheritance here

moradology: Does `TypedDict` act funny down the inheritance chain?

geospatial-jeff: No but it doesn't support `isinstance` or `issubclass` checks so I think the only point is DRY which seems unnecessary for only a few fields.

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/app.py
```diff
@@ -43,7 +45,7 @@ def run():
             host=settings.app_host,
             port=settings.app_port,
             log_level="info",
-            reload=settings.reload,
+            reload="true",
```
geospatial-jeff, line 48:
```suggestion
            reload=settings.reload,
```

stac_fastapi/pgstac/stac_fastapi/pgstac/app.py
```diff
@@ -17,13 +19,15 @@
 api = StacApi(
     settings=settings,
     extensions=[
-        TransactionExtension(client=TransactionsClient),
+        # TODO: Transaction extension should receive an instance of the class, not the class itself.
```
geospatial-jeff, line 22:
```suggestion
```

outdated

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -2,93 +2,36 @@
...
 from stac_fastapi.pgstac.types.search import PgstacSearch
-from stac_fastapi.types.core import BaseCoreClient
+from stac_fastapi.types.core import AsyncBaseCoreClient
 from stac_fastapi.types.errors import NotFoundError
+from stac_fastapi.types.stac import Collection, Conformance, Item, ItemCollection
 
 NumType = Union[float, int]
 
 
+# TODO: Add response handler instead of returning ORJSON classes
```
geospatial-jeff, line 20:
```suggestion
```
outdated

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -2,93 +2,36 @@
...
-from stac_pydantic import Collection, Item, ItemCollection
-from stac_pydantic.api import ConformanceClasses, LandingPage
-from stac_pydantic.shared import Link, MimeTypes, Relations
 
 from stac_fastapi.pgstac.models.links import CollectionLinks, ItemLinks, PagingLinks
 from stac_fastapi.pgstac.types.search import PgstacSearch
-from stac_fastapi.types.core import BaseCoreClient
+from stac_fastapi.types.core import AsyncBaseCoreClient
 from stac_fastapi.types.errors import NotFoundError
+from stac_fastapi.types.stac import Collection, Conformance, Item, ItemCollection
```
geospatial-jeff, line 15: do `from stac_fastapi.types import stac as stac_types` instead

geospatial-jeff:
Barring one failing test case in pgstac this PR is mostly done.  The initial goal of this PR was to better support custom data models by removing pydantic from the interface between API/backend.  This ended up being quite a lot more than I initially expected, here is a summary of the major changes:

# Code Changes

### 1. Replacing pydantic.
- Define `LandingPage`, `Conformance`, `Item`, `Catalog`, `Collection`, and `ItemCollection` types using `TypedDict`, which is just a more descriptive type annotation for `Dict[str, Any]`.
- Replace all calls to `stac_pydantic` in the base classes with the new `TypedDict` types.
- Remove `pydantic` from link factories, these just return dictionaries now.
- Refactor sqlalchemy and pgstac implementations of the base classes to use the new `TypedDict` types.  This mostly required switching from dot notation (`item.id`) to dictionary syntax (`item['id'])`.  But there were a few backend specific changes:
    - **sqlalchemy**
        - Removed pydantic orm mode and getterdicts.  Defined a new `Serializer` abstract base class which implements `Serializer.db_to_stac` and `Serializer.stac_do_db` methods for transforming between stac types and the database model.  An instance of the serializer is passed to the sqlalchemy backend for collections and items.
    - **pgstac**
        - Add `AsyncBaseCoreClient` which is identical to `BaseCoreClient` but implements `async def` functions instead of normal `def`.  Update the pgstac backend to use `AsyncBaseCoreClient`.  Also add a `AsyncBaseTransactionsClient` for the same reson.
        - Implement a `LandingPageMixin` class which is used by both `BaseCoreClient` and `AsyncBaseCoreClient` to create the initial landing page.
        - Update `pgstac` backend to return `TypedDict` instead of `ORJSONResponse` to match the base clients.
        - Lots of little changes to the pgstac transaction extension implementation to align with the base class.

### 2. API layer changes
I needed to make several changes to the API layer to accommodate the switch away from pydantic:
- Add the `ApiSettings.enable_response_models` attribute which determines whether or not API responses are validated against the appropriate `stac_pydantic` model (defaults to `false`).  When set to `false`, the `/docs` endpoint will not autogenerate OpenAPI docs for most endpoints.
- Add the `StacApi.response_class` attribute (defaults to `JSONResponse`) which determines what Starlette response class is used to send data back to the caller.  Defaults to `ORJSONResponse` for the pgstac implementation.
- Add the `StacApi.router` attribute which contains the FastAPI router used to build the application, defaults to a standard `APIRouter`.
- Changes to endpoint creation and registration.
    - Differentiation between `async def` and `def` endpoints is now determined based on an `isinstance` check against the backend client (ex. `isinstance(client, AsyncBaseCoreClient)`).
    - Added an additional `else` statement to endpoint factories to handle creating an endpoint with a `TypedDict`.  More on this below.
    - Define a single function for each route that is registered, making it easier for the caller to override route-level kwargs for a single route.

# Current Pydantic Usage
After all of these changes, where is pydantic still used?
- Validating the input to `POST /search` requests.
- For settings management in `ApiSettings` (`pydantic.BaseSettings`).
- Validating API responses if `ApiSettings.enable_response_models = True`.
- In the `tiles` extension (integration with titiler/ogc tiles).  Both titiler/ogc tiles are pretty rigid so it makes sense to use pydantic here.
- In test cases to help with dictionary equality comparisons.
- Sqlalchemy backend still uses pydantic includes/excludes syntax to implement the fields extension.

# Futher Consideration
There are several things which need further consideration:
1. Pydantic has support for validating `TypedDict` (https://pydantic-docs.helpmanual.io/usage/types/#typeddict), which means if we specify a `TypedDict` as the request model of a FastAPI api route it will create a pydantic model matching the `TypedDict` which is used to validate the API responses.  This breaks the API/backend interface because the backend is expecting a dictionary but the API gives it a pydantic model.  For now the route factories fall back to using the `Dict[str, Any]` type declaration instead of the appropriate `TypedDict` to side step this behavior, but it might be a useful feature.  Either way we need to spend more time looking into this interaction.
2. `TypedDict` is a pretty new feature in python, and doesn't properly differentiate between required and optional field suntil python3.9.  The `typing_extensions` package does have this differentiation, so we should consider switching to `typing_extensions`.
3. Endpoint factories are quite flimsy, we really want them to be transparently wrapping the underlying function but instead we are brute forcing it with a bunch of `isinstance` checks and conditional logic.  It "works" for now but I won't be surprised if it breaks in the future.  I'm thinking a combination of decorators + type introspection could fix this nicely.
4. There is a flag which enables using pydantic for response models, it may be valuable to users to also provide a flag for enabling pydantic for request models as well.  Especially for the transactions extension where the backend code is relying on the data to be structured a certain way.  But for now this is not implemented.

# Bugs
I found a few other bugs that aren't really related to this PR:

### pgstac
1. The `TransactionsClient.delete_item` and `TransactionsClient.delete_collection` methods expect to return a `stac_types.Item` class, but pgstac does not include this information which causes the implementation to diverge from what the base class expects.  This isn't a big deal because the delete functionality still works, but I thought I would mention it.  The probable solution here is to update the base class to return `None` on delete endpoints which is just as RESTful as returning the thing being deleted.
2. When using the fields extension with pgstac, if `item.collection` is not returned it will cause the code to fail because the pgstac search implementation (`_search_base`) atempts to access this information (`item['collection']`) when building links.  For now I've updated the pgstac fields extension test cases to always return `item.collection`.

### sqlalchemy
1. This backend currently won't return `stac_extensions`.  I can't find it in my notes and can't remember why, but there was a good reason for this at the time lol.  Shouldn't be too hard to fix.
2. We are using a custom Geoalchemy2 geometry type which converts the geometry to geojson upon access, but this has become annoying because this conversion only happens if we are accessing an orm object bound to the database.  If we call `item.geometry` on an unbound orm object it returns a `WKBElement`.  This type divergence is annoying, probably best to remove the custom geometry type.

geospatial-jeff:
@bitner The `tests/resources/test_item.py::test_pagination_post` pgstac test case is failing, the test requests 5 items with `limit=1` but only ever finds 3 pages.  I've looked at my changes and I don't think I changed the underlying pagination logic, could you please take a look?

bitner:
Hey @geospatial-jeff still looking through things and checking out the test case with pagination issue, but one thing I noticed: @lossyrob had made the observation that Pydantic type checking in many places where it was redundant on data that is known and "should" already be validated was causing a very noticeable slow down. That is why we had switched to just directly outputing ORJSON rather than Pydantic models and also why when we were using Pydantic models essentially to pass through data we were using model.construct(**foo) rather than model(**foo).

geospatial-jeff:
> That is why we had switched to just directly outputing ORJSON rather than Pydantic models and also why when we were using Pydantic models essentially to pass through data we were using model.construct(**foo) rather than model(**foo).

This hasn't changed with this PR.  Pgstac backend returns `TypedDict` (not pydantic) and uses an `ORJSONResponse` as the route's `response_class`.  So if anything it should be slightly faster now because we are skipping any calls to pydantic by default.

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -174,32 +114,32 @@ async def _search_base(
...
+        collection = ItemCollection(**items)
         cleaned_features = []
-        if collection.features is None or len(collection.features) == 0:
+        if collection["features"] is None or len(collection["features"]) == 0:
             raise NotFoundError("No features found")
 
-        for feature in collection.features:
-            feature = Item.construct(**feature)
+        for feature in collection["features"]:
+            feature = Item(**feature)
```
bitner, line 123: @geospatial-jeff here is an example of one of the places that we were using Item.construct(**feature) as opposed to Item(**feature) to avoid the pydantic overhead.

geospatial-jeff:
Oh as I'm reading more I realize the `response_class` is just there for documentation, I'll fix this today by directly returning response classes in the endpoint factories.

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
geospatial-jeff: in this case `Item` is not a pydantic model.  It's just a `TypedDict`.

lossyrob:
@geospatial-jeff thanks for the write up - this looks like a set of great changes. I'm going to try and review this this week.

One question I had which is related to the pydantic validation stuff - you mention that `enable_response_models = True` would validate the response models, which I'm assuming means passing them through pydantic validation? In that case for the optimization David's mentioning, we'd want to have that be False so that the response goes directly from TypedDict -> ORJsonResponse. So seems like we'd want this to be False. But is there any way to get the OpenAPI docs to document responses with Pydantic models (in case we'd like that feature), without passing the responses through Pydantic validation?

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
lossyrob: I think in this case, this is no longer a Pydantic model but a TypedDict instead (https://github.com/stac-utils/stac-fastapi/blob/custom-models/stac_fastapi/types/stac_fastapi/types/stac.py#L50), so it's creating a different type of object

bitner: ahh, got it. missed that part.

geospatial-jeff: I'm thinking of renaming the types to `ItemDict`, `CollectionDict` etc. to be more explicit.  Too many classes named `Item` etc.

jonhealy1:
These changes look really good

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -79,100 +84,200 @@ def get_extension(self, extension: Type[ApiExtension]) -> Optional[ApiExtension]
...
+        """Register get item collection endpoint (GET /collection/{collectionId}/items).
+
+        Returns:
+            None
+        """
+        self.router.add_api_route(
             name="Get ItemCollection",
             path="/collections/{collectionId}/items",
-            response_model=ItemCollection,
+            response_model=ItemCollection
```
lossyrob, line 244: Would be a nice feature to override the pydantic response_model via config like [this draft PR was doing](https://github.com/stac-utils/stac-fastapi/pull/134/files#diff-4f8c5fdc9b3cca78110705589c68a6df26ac7a9d66bd4af3108db5174fdfe49cR139), but I think this can be done in a follow-up. Reminder to self to create an issue.

lossyrob:
Seeing a bit of a downside to TypedDict in situations where things are optionally included. E.g. https://github.com/stac-utils/stac-fastapi/blob/custom-models/stac_fastapi/types/stac_fastapi/types/stac.py#L47 - it would be nice to have this as an `Optional[Dict[str, Any]]`, and if were `None` to make sure it's not included in the JSON response (like pydantic's `dict(exlude_none=True)`. There's not a good mechanism for that. You mention in your notes that potentially 3.9 or typing_extensions handle optional attributes better, but I don't see where that is - I only see [PEP 655 - Marking individual TypedDict items as required or potentially-missing](https://www.python.org/dev/peps/pep-0655/), which is still a draft.

Another instance of this is the attribute `stac_extensions: Optional[List[str]]` - you actually wouldn't want this to be set as `None`, as that would translate to having `"stac_extensions":  null,` in the JSON output, which I don't believe would be valid STAC. I'm likewise going to comment it out because of this issue.

This generally makes me question the utility of TypedDict as a benefit over Dict[str, Any]...the annotated types won't be complete and therefore may cause more confusion than anything. As a way forward, I'm going to document these additional fields in the `types.stac` types, but not include them as TypedDict entries. That way we can avoid having nulls, and if they are included in the construction of the TypedDict, that works fine - though setting them explicitly causes a type checking error.

geospatial-jeff:
`TypedDict` supports differentiation between required/optional fields in python3.9.  But the `typing_extensions.TypedDict` type already supports this.  I mentioned this in a previous comment:

> TypedDict is a pretty new feature in python, and doesn't properly differentiate between required and optional field suntil python3.9. The typing_extensions package does have this differentiation, so we should consider switching to typing_extensions.

So I'd recommend just switching to using `typing_extensions` until we decide to switch to python3.9.  I think the value of using `TypedDict` over `Dict[str, Any]` is it makes it much clearer what the expected keys are (and their types) which should make it easier for people who are unfamiliar with the STAC spec to build their own backend, or understand how to access certain dictionary members in the responses of existing backends.

I'm fine with switching to `Dict[str, Any]` as well.

geospatial-jeff:
> You mention in your notes that potentially 3.9 or typing_extensions handle optional attributes better, but I don't see where that is - I only see PEP 655 - Marking individual TypedDict items as required or potentially-missing, which is still a draft.

Oh I totally read over this.  I pulled this from the pydantic docs.  But I guess the release slipped to 3.10.  So probably best to use `Dict[str, Any]` for now.

lossyrob:
I saw you mentioned it, and mentioned that:
> You mention in your notes that potentially 3.9 or typing_extensions handle optional attributes better, but I don't see where that is - I only see  [PEP 655 - Marking individual TypedDict items as required or potentially-missing](https://www.python.org/dev/peps/pep-0655/) - Marking individual TypedDict items as required or potentially-missing, which is still a draft.

</mention inception>

Can you link to docs around the optional fields in python 3.9 (I'm only seeing [these docs](https://docs.python.org/3/library/typing.html#typing.TypedDict) for 3.9.6 which doesn't mention it)? For the typing_extensions implementation, I'm seeing [this code](https://github.com/python/typing/blob/master/typing_extensions/src_py3/typing_extensions.py#L1722) but not docs on how it's used or other indications about that feature. Do you have any links about how to utilize the optional and required fields?

Another option I thought about is to have clients return either a Dict[str, Any] _or_ a pydantic model (or perhaps a PySTAC object for that matter). That way if you want the clear development path as you mention, you can use those guiding tools, but if you want the speed of just dealing with Dicts, you can also go that route. Would be matter of changing signatures like

```
def get_collection(self, id: str, **kwargs) -> Collection:
```

to

```
def get_collection(self, id: str, **kwargs) -> Union[Collection, Dict[str, Any]]:
```

Where `Collection` can either be the TypedDict, from stac-pydantic or from PySTAC.  What do you think?

geospatial-jeff:
I really hate Union types, they are a PITA.  I'd rather just have a single return type.

I'd say for this PR lets go with `Dict[str, Any]` because it requires the least change, and maybe we can open an issue for switching to `pystac` types because I think that might be a good thing long term.

lossyrob:
> I'd say for this PR lets go with Dict[str, Any] because it requires the least change, and maybe we can open an issue for switching to pystac types because I think that might be a good thing long term.

Sounds like a plan.

The thing about PySTAC types is there's some copying in and out of dicts that I'd rather avoid along critical paths. @duckontheweb had mentioned an idea to make all PySTAC types wrap dicts of properties rather then translate in and out of dicts at construction time - this is something we do a lot in the extensions. If we did this, then the cost of serialization/deserialization pretty much goes away, in which case I think using PySTAC here might be a good option. Not sure if that's generally the direction to go in PySTAC, but this is certainly a data point in the +1 column.

I'll change things over to use `Dict[str, Any]` and push things up to this branch; I've merged in the latest and fixed most tests and I think it's pretty close.

lossyrob:
Ended up keeping `TypedDict`, but using `total=False`; this makes all of the attributes optional, which gives the benefit of a simple way to check what STAC properties are expected, but without enforcing any potentially incorrect requirements.

geospatial-jeff:
@lossyrob your commits look great!  I can't approve the PR since I opened it but this is ready to merge.  Once it's merged I'm going to do a follow up PR to add documentation for some of the new features we've introduced.

lossyrob (approved):

Issue #159, opened 2021-06-17
Add conformance classes to Landing Page


lossyrob:
The LandingPage should include conformance classes to describe the conformance classes, in addition to the conformance endpoint.

In addition, there's currently a single conformance class. I'd recommend using:

```
"conformsTo": [
"https://api.stacspec.org/v1.0.0-beta.1/core",
  "https://api.stacspec.org/v1.0.0-beta.1/item-search",
  "http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core",
  "http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/oas30",
  "http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/geojson"
]
```

Along with adding conformance classes based on what extensions are enabled.

From: https://github.com/radiantearth/stac-api-spec/tree/master/core

> The conformsTo section must provide the capabilities of this service. This is the field that indicates to clients that this is a STAC API and how to access conformance classes, including this one. The relevant conformance URI's are listed in each part of the API specification. If a conformance URI is listed then the service must implement all of the required capabilities.

philvarner:
Does stac-fastapi not implement OAFeat also? if so, it should include https://api.stacspec.org/v1.0.0-beta.2/ogcapi-features

PR #165, opened 2021-06-30
Add filter extension

rsmith013:
Adding filter extension. Also making changes to other extensions to make it easier to dynamically build the conformance classes from the implemented extensions.


stac_fastapi/extensions/stac_fastapi/extensions/core/filter.py
```diff
@@ -0,0 +1,63 @@
...
+
+from stac_fastapi.api.models import CollectionUri, EmptyRequest
+from stac_fastapi.api.routes import create_endpoint
+from stac_fastapi.types.core import BaseFiltersClient
+from stac_fastapi.types.extension import ApiExtension
+
+
+@attr.s
+class FilterExtension(ApiExtension):
+    """Transaction Extension.
```
matprov, line 17: Replace with actual extension name

rsmith013: Done

stac_fastapi/extensions/stac_fastapi/extensions/core/fields.py
```diff
@@ -37,6 +38,10 @@ class FieldsExtension(ApiExtension):
         )
     )
 
+    conformance_classes: List[str] = attr.ib(
```
geospatial-jeff, line 41: These should be class variables, as they won't change between instances of the class.

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py
```diff
@@ -89,7 +88,7 @@ def item_collection(
                 .order_by(self.item_table.datetime.desc(), self.item_table.id)
             )
             count = None
-            if self.extension_is_enabled(ContextExtension):
+            if self.extension_is_enabled("ContextExtension"):
```
geospatial-jeff, line 91: What's the reasoning behind this change?

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -124,9 +124,22 @@ class BaseCoreClient(abc.ABC):
...
 
-    def extension_is_enabled(self, extension: Type[ApiExtension]) -> bool:
+    def extension_is_enabled(self, extension: str) -> bool:
         """Check if an api extension is enabled."""
-        return any([isinstance(ext, extension) for ext in self.extensions])
+        return any([type(ext).__name__ == extension for ext in self.extensions])
+
+    def list_conformance_classes(self):
+        """Return a list of conformance classes, including implemented extensions."""
+        base_conformance = [
```
geospatial-jeff, line 133: I believe this is the set of conformance classes we want to use https://github.com/stac-utils/stac-fastapi/issues/159

geospatial-jeff:
This also closes #159 by adding conformance classes to the landing page.

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py
rsmith013: This change links somewhat to the thinking which spawned #178. It would be really good to try and un-bake the extensions from the core client. The reasoning behind this change is to remove the need to import the Extension classes to check if the extension is active. The client instance has the Extension instances passed to it at runtime so I changed the `is_extension_enabled` method to enable simple name checking https://github.com/rsmith013/stac-fastapi/blob/247098fbb28345096080ea161158ce4461868abb/stac_fastapi/types/stac_fastapi/types/core.py#L127

stac_fastapi/extensions/stac_fastapi/extensions/core/filter.py
```diff
@@ -0,0 +1,63 @@
...
+        client: Queryables endpoint logic
+        conformance_classes: Conformance classes provided by the extension
+
+    """
+
+    client: BaseFiltersClient = attr.ib()
+    conformance_classes: List[str] = attr.ib(
+        default=[
+            "https://api.stacspec.org/v1.0.0-beta.2/item-search#filter",
+            "https://api.stacspec.org/v1.0.0-beta.2/item-search#filter:simple-cql",
```
philvarner, line 36: FYI, these are about to change. https://github.com/radiantearth/stac-api-spec/pull/163

stac_fastapi/types/stac_fastapi/types/core.py
philvarner: yes, these are the old conformance classes here. #159 has the most up to date ones.

stac_fastapi/extensions/stac_fastapi/extensions/core/fields.py
rsmith013: While that is true in most cases, it is not in all. The filter extension is a good example. As the filter extension is complex, there are varying degrees of implementation. There is a base level to be compliant but then there are other conformance classes that are optional and implementation-specific. Having these as instance variables allows the implementor to specify which classes their implementation conforms to.

https://github.com/radiantearth/stac-api-spec/tree/master/fragments/filter#conformance-classes

stac_fastapi/types/stac_fastapi/types/core.py
rsmith013: presumably we are on beta2 now. 

e.g. https://api.stacspec.org/v1.0.0-beta.2/core

Offically 1.0.0 has been released but it looks like the openapi schemas are not there yet.

stac_fastapi/extensions/stac_fastapi/extensions/core/filter.py
rsmith013: Have changed the conformance classes to:

```python
        default=[
            "https://api.stacspec.org/v1.0.0-beta.2/item-search#filter:basic-cql",
            "https://api.stacspec.org/v1.0.0-beta.2/item-search#filter:item-search-filter",
            "https://api.stacspec.org/v1.0.0-beta.2/item-search#filter:cql-text",
        ]
```

this was my interpretation of the PR where the basic spatial and temporal operators were recommended. I added `cql-text` as default, as item-search must implement `GET` search. It would follow that the base implementation of the filters extension would also be a `GET` operation using `cql-text`

stac_fastapi/types/stac_fastapi/types/core.py
philvarner: Correct.

stac_fastapi/extensions/stac_fastapi/extensions/core/filter.py
philvarner: Sorry, I was actually just trying to let you know that they were going to change for beta.3. For beta.2, it should still use `simple-cql`

rsmith013: oh, ok. will revert.

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -141,6 +157,7 @@ def landing_page(self, **kwargs) -> LandingPage:
             id=self.landing_page_id,
             title=self.title,
             description=self.description,
+            conformsTo=self.list_conformance_classes(),
```
philvarner, line 160: This looks like it will fix https://github.com/stac-utils/stac-fastapi/issues/169

philvarner: @rsmith013  I thought this would fix #169, but when I ran this using docker-compose from your branch, the landing page `conformsTo` is still empty. maybe i'm not running it right? This is a pretty major bug in stac-fastapi, so I just wanted to make sure it gets fixed. Do you see `conformsTo` in the landing page when you run it locally?

rsmith013: yes:
image

philvarner:
Thanks for checking! I realized later that my venv somehow had the
stac-fastapi packages installed in it, so I think those were being used
instead of the source.

lossyrob:
Making sure I understand - this PR adds a filter extension that implements the queryables routes, but does not implement any of CQL "filter" property on the `Search` object, correct? So it's up to backends to implement the actual CQL filtering, but this PR adds the necessary top level routes that are added by the extension? If I'm getting that correct, this seems like a good change that will set up the next stage of work which is to add CQL support to the pgstac and sqlalchemy backends.

@rsmith013  Can you merge in the latest master, and I'll take a closer look next week? Thanks!

rsmith013:
@lossyrob Yes, that was the plan.

lossyrob:
Looking good; going to leave this off of the 2.0.0 release because there's a TODO around implementing the Async core client.

rsmith013:
@lossyrob I didn't realise you were that close to a release. Added the Async base and merged master. Should be good to go

lossyrob:
Thanks! We decided to pop one off since it had been so long. Let use kick the tires for the CI-based publishing as well. There's a few issues so I'm sure we're going to have to fix up soon so I'm expecting a 2.1 release in the next week or so

stac_fastapi/types/stac_fastapi/types/core.py
moradology: Recent updates to the library have pushed things from 1.0.0-beta.2 to 1.0.0 in case that was missed

stac_fastapi/types/stac_fastapi/types/core.py
duckontheweb: @philvarner You may need to rebuild the images when running `docker-compose` to see these changes:

```console
$ docker-compose up --build ...
```

lossyrob (approved):

PR #180, opened 2021-07-06
Retry on connection failure during joplin ingest

### LINKED ISSUE #170: errors using docker compose commands to start servers
Per the instructions, running

```
docker-compose build
docker-compose up
```

does not seem to work. I can connect to the pgstac server on 8082, but not the sqlalchemy one on 8081, and the sqlalchemy server seems to be constantly restarting. 

I see these errors in the startup log:

```
pgstac-loadjoplin_1     | requests.exceptions.ConnectionError: HTTPConnectionPool(host='app-pgstac', port=8082): Max retries exceeded with url: /collections (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f69d91529a0>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

```
sqlalchemy-migration_1  | requests.exceptions.ConnectionError: HTTPConnectionPool(host='app-sqlalchemy', port=8081): Max retries exceeded with url: /collections (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f09e1e4e9d0>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

This is using Docker Desktop for MacOSX.

duckontheweb:
Fixes #170

### DIFF
--- a/scripts/ingest_joplin.py
+++ b/scripts/ingest_joplin.py
@@ -5,11 +5,20 @@ from pathlib import Path
 from urllib.parse import urljoin
 
 import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 workingdir = Path(__file__).parent.absolute()
 joplindata = workingdir.parent / "stac_fastapi" / "testdata" / "joplin"
 
 app_host = sys.argv[1]
+
+retry_strategy = Retry(connect=5, backoff_factor=2)
+adapter = HTTPAdapter(max_retries=retry_strategy)
+
+session = requests.Session()
+session.mount("https://", adapter)
+session.mount("http://", adapter)
 
 if not app_host:
     raise Exception("You must include full path/port to stac instance")
@@ -21,7 +30,7 @@ def ingest_joplin_data(app_host: str = app_host, data_dir: Path = joplindata):
     with open(data_dir / "collection.json") as f:
         collection = json.load(f)
 
-    r = requests.post(urljoin(app_host, "collections"), json=collection)
+    r = session.post(urljoin(app_host, "collections"), json=collection)
     if r.status_code not in (200, 409):
         r.raise_for_status()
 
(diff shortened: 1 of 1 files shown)


scripts/ingest_joplin.py
```diff
@@ -5,12 +5,21 @@
...
 joplindata = workingdir.parent / "stac_fastapi" / "testdata" / "joplin"
 
 app_host = sys.argv[1]
 
+retry_strategy = Retry(connect=5, backoff_factor=2)
+adapter = HTTPAdapter(max_retries=retry_strategy)
+
+session = requests.Session()
+session.mount("https://", adapter)
+session.mount("http://", adapter)
```
kylebarron, line 21: 🤯 That's so cool!

geospatial-jeff:
I'm guessing the root cause of the issue is the database/application containers becoming available after the ingest script is run?

geospatial-jeff (approved):

duckontheweb:
> I'm guessing the root cause of the issue is the database/application containers becoming available after the ingest script is run? Adding retry/backoff will fix this.

Yeah, the containers themselves are available, but the FastAPI app within the container must still be booting up in some cases. Maybe the better way to address this is actually using a [`healthcheck`](https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck). I'll put in a separate PR with that approach to see how it compares.

geospatial-jeff:
> actually using a healthcheck

I've used this tool before (https://github.com/vishnubob/wait-for-it) in k8s deployments when one container needs to wait for another, could work here.

duckontheweb:
> > actually using a healthcheck
> 
> I've used this tool before (https://github.com/vishnubob/wait-for-it) in k8s deployments when one container needs to wait for another, could work here.

Opened #182 using `wait-for-it.sh` and I like that approach a lot better. Thanks for suggesting that!

duckontheweb:
Closing in favor of #182

PR #181, opened 2021-07-07
Update to stac-pydantic 2.0.0

moradology:
This PR upgrades the version of `stac-pydantic` to 2.0.0 so that the library can use stac-spec 1.0.0. Mostly, the changes are import updates, though collections are now represented as `stac-pydantic`'s `Collections` type rather than a `List[Collection]`

### DIFF
--- a/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
+++ b/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
@@ -39,6 +39,7 @@ class TransactionsClient(BaseTransactionsClient):
     ) -> schemas.Collection:
         """Create collection."""
         data = self.collection_table.from_schema(model)
+        data.type = "collection"
         with self.session.writer.context_session() as session:
             session.add(data)
             data.base_url = str(kwargs["request"].base_url)
--- a/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
+++ b/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
@@ -39,7 +39,6 @@ class TransactionsClient(BaseTransactionsClient):
     ) -> schemas.Collection:
         """Create collection."""
         data = self.collection_table.from_schema(model)
-        data.type = "collection"
         with self.session.writer.context_session() as session:
             session.add(data)
             data.base_url = str(kwargs["request"].base_url)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+## UNRELEASED
+
+* Upgrade to stac-pydantic 2.0.0 and stac-spec 1.0.0 (https://github.com/stac-utils/stac-fastapi/pull/181)
 
 ## 1.1.0 (2021-01-28)
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -92,7 +92,7 @@ services:
       - ./stac_fastapi:/app/stac_fastapi
       - ./scripts:/app/scripts
     command: >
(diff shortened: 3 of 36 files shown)


moradology:
Apologies if the failed linting is blowing up anyone's emails - I'm attempting to track down the options I'm failing to use which seem to be expected by CI's run of `black`

moradology:
It looks as though stac-pydantic likes `black-19.10b0` whereas stac-fastapi is a fan of `black-21.6b0`

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
```diff
@@ -39,6 +39,7 @@ def create_collection(
     ) -> schemas.Collection:
         """Create collection."""
         data = self.collection_table.from_schema(model)
+        data.type = "collection"
```
moradology, line 42: This is most certainly not going to be kept. It ought to be removed before the PR is accepted as it is residue from debugging

moradology:
There's one test that is failing and which I'm having a hard time making heads or tails of: https://github.com/stac-utils/stac-fastapi/blob/a568d871396c65c7a514bb7d2dfbb669798a17ae/stac_fastapi/sqlalchemy/tests/resources/test_item.py#L163-L189. A second or third pair of eyes might be useful at this point...

At nearly every step along the request/response cycle that I can surface the `created` timestamp, it appears to be correctly formatted (e.g. https://github.com/stac-utils/stac-fastapi/blob/a568d871396c65c7a514bb7d2dfbb669798a17ae/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/models/database.py#L106-L108). It is also the case that the appropriate pydantic serialization arguments have been provided via `class Config`. Nevertheless, whenever that field is inspected in the `test_item_timestamps` test, RFC 3339 formatting (which we expect) is not there.

geospatial-jeff:
@moradology I don't know what is going on, but it's related to how FastAPI is applying the `response_model`.  For some reason it is adding a time zone component even though we aren't telling it to do that, although this is still RFC3339 compliant so maybe its not all that bad.  The test failure is happening because the pydantic model is using pydantic's datetime parser while the test itself is expecting a very specific type of RFC3339 string coming from stac-pydantic.  We can fix the test by updating the test case to use pydantic's datetime parser.

I'm still confused why adding a `response_model` changes how the datetime is serialized.  It seems to be specific to context, if you call `model.json()` or `model.dict()` we get a date string that ends with `Z` but when FastAPI applies the response_model it creates a date string with a timezone ending in `+00:00`.  Smells very funky, but maybe its not an issue because the date included in the response is still valid RFC3339.

geospatial-jeff:
Thinking through this a little more:
1. We are using `stac_pydantic.shared.DATETIME_RFC3339` in other test cases and those tests don't break.  Which means this seems to be specific to updating existing items in the database.  No idea why this is the case 🤷‍♂️ 
2. This tells me that it's probably related to the database or sqlalchemy update syntax, the database is using timestamps but if this were really the cause of the underlying problem I think we'd see the other test cases fail.
3. This is even more reason to move away from using pydantic for response models (ref #147).

Either way I've fixed this test in https://github.com/stac-utils/stac-fastapi/pull/181/commits/bfa203ead6452556adeac698b8c67097dad8cdc6

geospatial-jeff:
@moradology I'm marking this ready for review, please take a look over the changes I made.

lossyrob:
@geospatial-jeff thinking through review strategy, which I have some time set aside for today - I'd like to review this PR and get it merged, and then update #147 with these changes and integrate while reviewing. I'll make tweaks along the way but will tag you for any larger review changes if I find them. Let me know if you want to tweak that plan, otherwise I'm going to get started.

geospatial-jeff:
That sounds good to me.  Either way the merge conflicts are going to be nasty :)

lossyrob (approved):
LGTM. Added a couple of commits on top that I needed for testing the PR in the dockerized environment. Will merge once this passes and integrate with #147

PR #184, opened 2021-07-08
io specific mods

aliasmrchips:
Change geometry field from Polygon to Geometry to allow MultiPolygons (in particular)

Add queryable fields and types


lossyrob:
There's a lot of use-case specific changes here, and I don't think this applies to the generalized STAC API. Several users have struggled with customizing the sqlalchemy backend without having to fork and make changes to the schema. We did a similar mod before moving to the pgstac backend, which doesn't have the schema restrictions that the sqlalchemy backend does, and I'd recommend giving that backend a look. But unless there's generalized improvements to the sqlalchemy backend contained here, I'd recommend that stac-fastapi not merge in PRs with use-case specific fields.

aliasmrchips:
True, I did make this branch to keep track of any changes we made... However, I did not intend to create a PR and don't remember doing so. At any rate, we will likely be migrating to the pgstac backend as you suggest.

PR #188, opened 2021-07-20
use setup.cfg metadata to set package version

vincentsarago:
This PR removes the code in the setup.py which grab the `version` form the code and use setup.cfg metadata notation

This is a personal opinion so feel free to ignore. (I also had difficulties trying to install the package using GitHub url, but not 💯 it was related to this)


geospatial-jeff (approved):

vincentsarago:
can we merge this ?

Issue #192, opened 2021-07-26
enable async bulk endpoint 


vincentsarago:
https://github.com/stac-utils/stac-fastapi/blob/50605015ec72043309bfea5163d753fab6358d77/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py#L83-L85

FYI: I'm working on a pgstac bulk loader ;-)

PR #193, opened 2021-07-26
[DO NOT MERGE] add bulkTransaction for pgstac

vincentsarago:
add bulk transaction for pgstac

This also introduce a new function `stac_fastapi.api.routes.create_endpoint`(related to #192) which either create an async or a sync endpoint).

DO NOT MERGE, this is linked to https://github.com/stac-utils/pgstac/pull/32 and thus to the next release of pgstac

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/routes.py
+++ b/stac_fastapi/api/stac_fastapi/api/routes.py
@@ -1,4 +1,5 @@
 """route factories."""
+import asyncio
 from typing import Any, Callable, Dict, Type, Union
 
 from fastapi import Depends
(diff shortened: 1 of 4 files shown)


stac_fastapi/api/stac_fastapi/api/routes.py
```diff
@@ -94,3 +95,15 @@ def _endpoint(
...
+def create_endpoint(
+    func: Callable,
+    request_model: Union[Type[APIRequest], Type[BaseModel], Dict],
+    response_class: Type[Response] = JSONResponse,
+):
+    """Create FastAPI Async or Sync endpoint."""
+    if asyncio.iscoroutinefunction:
+        return create_async_endpoint(func, request_model, response_class)
+    else:
+        return create_sync_endpoint(func, request_model, response_class)
```
vincentsarago, line 109: I'm happy to update other code to use this function if we think it's 👌

vincentsarago:
let's re-work that later

PR #199, opened 2021-07-27
Preserve relative paths on link generation

moradology:
This PR addresses a couple of small issues introduced with the recent stac-pydantic 2.0 and model customization updates. First, `urljoin`s now build links preserving relative paths. Second, conformance classes are now on `CoreCrudClient` and its async variant, enabling users to customize these values when instantiating a client


lossyrob:
Thanks for catching. Can we add tests to make sure we catch this in the future if it regresses again?

moradology:
I believe this is ready for a once over

geospatial-jeff (approved):
LGTM!

PR #200, opened 2021-07-28
add in urljoin to complete conformance href

davidraleigh:
landing page fix to complete conformance href

before:
before-conformance

after conformance-url fix:
after-conformance


davidraleigh:
fixed in https://github.com/stac-utils/stac-fastapi/pull/199

Issue #206, opened 2021-08-05
stac-fastapi-sqlalchemy docker-compose execution fails


philvarner:
Running per the README:

```
docker-compose build
docker-compose up
```

then retrieving http://localhost:8081/

results in:

```
{
  "detail": "unhandled database error"
}
```

bitner:
I believe this is the same issue as #205

philvarner:
works for me.

Issue #205, opened 2021-08-05
Possible breaking change in docker-compose.yml


jaysnm:
Hi

Change of database image from `bitner/pgstac:0.2.4` to `ghcr.io/stac-utils/pgstac:v0.2.8` breaks system spin with 
`ERROR: Head https://ghcr.io/v2/stac-utils/pgstac/manifests/v0.2.8: denied: denied`.
However, reversing image back to `bitner/pgstac:0.2.4` or `bitner/pgstac:0.2.7` (latest) works fine.

geospatial-jeff:
@bitner any ideas here?

bitner:
@geospatial-jeff I'll look

jaysnm:
Turns out to be `ghcr authorization` related problem. Works fine after a `docker logout ghcr.io`. See #208.

PR #208, opened 2021-08-06
Wait for 5432 to be available before running apps in docker

bitner:
Use wait-for-it.sh on port 5432 (Postgres) before starting the app in both sqlalchemy and pgstac backends on Docker.

I believe this is what was causing the issues in #205 and #206

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,7 +32,7 @@ services:
     depends_on:
       - database
     command:
-      python -m stac_fastapi.sqlalchemy.app
+      bash -c "./scripts/wait-for-it.sh database:5432 && python -m stac_fastapi.sqlalchemy.app"
 
   app-pgstac:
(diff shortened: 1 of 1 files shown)


jaysnm:
#205 looks like access denial to `ghcr.io/stac-utils/pgstac:v0.2.8` image. I tried docker pull, got `Error response from daemon: Head https://ghcr.io/v2/stac-utils/pgstac/manifests/v0.2.8: denied: denied`

kylebarron:
for another data point, I'm able to pull that image with no problem.

jaysnm:
Had to check with a remote host to verify it's an issue on my machine. Works fine after a `docker logout ghcr.io`

jaysnm (approved):
Works fine!

geospatial-jeff (approved):

PR #209, opened 2021-08-06
Add Support for CQL JSON to Stac FastAPI PGStac Backend

bitner:
Adds filter to PGStac Search to allow CQL filters to be passed to the PGStac backend to leverage the new CQL functionality in https://github.com/stac-utils/pgstac/pull/38


vincentsarago:
is this ready to go ?

PR #210, opened 2021-08-06
adding `item_serializer` and `item_table` to `BulkTransactionsClient`

davidraleigh:
I'd like to propose using the `item_table` and `item_serializer` that are already found in the `TransactionsClient` in `BulkTransactionsClient`.

This allows for inheritance from the `BulkTransactionsClient` class. That in turn allows for defining a custom database model (that itself can inherit from `stac_fastapi.sqlalchemy.model.database.Item`)


geospatial-jeff:
Agreed this is a good change, thanks for the PR!

geospatial-jeff (approved):
cheers @davidraleigh

Issue #207, opened 2021-08-06
Enable context extension for Search operations - search support for common STAC clients


jaysnm:
Many thanks to maintainers of this cool project. 
I'm not sure if this this a feature request or bug report, nonetheless, this is what I have noted on `/search` functionality

I have tried catalog search using both [sat-search](https://github.com/sat-utils/sat-search) and [pystac-client](https://github.com/stac-utils/pystac-client) but none seems to have native/compatible support.

While `sat-search` expects `/search` to return statistics of features matching query with no payload of feature items when `limit` is set to `0`, stac-fastapi responds with `404  Error: Not Found`

```
curl -X 'POST' \
  'http://162.10.0.1:9092/search' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
    "collections": ["ls8_sr_c2_l2"],
    "bbox": [-180, -90, 180, 90],
    "datetime": "2021-01-01/2021-08-01",
    "limit": 0
}'
```
Response
```
{
  "detail": "No features found"
}
```
Without `limit=0`
```
{
  "type": "FeatureCollection",
  "links": [
    {
      "rel": "next",
      "type": "application/json",
      "method": "POST",
      "href": "http://162.10.0.1:9092/search",
      "body": {
        "collections": [
          "ls8_sr_c2_l2"
        ],
        "bbox": [
          -180,
          -90,
          180,
          90
        ],
        "datetime": "2021-01-01/2021-08-01",
        "token": "next:c51cb097-e074-5768-bdcc-f1eeb1b6bd31"
      }
    },
    {
      "rel": "root",
      "type": "application/json",
      "href": "http://162.10.0.1:9092/"
    },
    {
      "rel": "self",
      "type": "application/json",
      "href": "http://162.10.0.1:9092/search"
    }
  ],
  "features": [
    {
      "id": "d0261e13-6fed-5c75-9dc4-5878aceae1cc",
  .....
}]
```

Executing a similar request on `https://earth-search.aws.element84.com/v0/search` with `limit=0` on `sentinel-s2-l1c` collection yields

```
{'type': 'FeatureCollection',
 'stac_version': '1.0.0-beta.2',
 'stac_extensions': [],
 'context': {'page': 1, 'limit': 0, 'matched': 1731850, 'returned': 0},
 'numberMatched': 1731850,
 'numberReturned': 0,
 'features': [],
 'links': [{'rel': 'next',
   'title': 'Next page of results',
[...]

PR #211, opened 2021-08-07
Enable context extension by default

jaysnm:
A work around issue #207 

`ContextExtension` for `pgstac` need be implemented, not working at the moment.


stac_fastapi/pgstac/stac_fastapi/pgstac/app.py
```diff
@@ -27,6 +28,7 @@
         QueryExtension(),
         SortExtension(),
         FieldsExtension(),
+        ContextExtension(),
```
lossyrob, line 31: pgstac does not currently support the context extension, and so shouldn't include it by default in the current implementation. However, the newest releases of pgstac do support this extension. Once we bump up the pgstac version we could potentially include this.

jaysnm: Thanks @lossyrob

The new version seems not to support context extension for pgstac. I have looked at implementation on sqlalchemy core.py and seems I can dirt hands on this. Let me know if you need an extra hand!

rsmith013:
I would think that the extensions shouldn't be enabled by default. I would hope that stac-fastapi becomes a framework that provides all the tools to accomplish what you need. I would think that it is down to the implementer to enable the context extension (and any other extension) that will be required by their community of users and for clients to inspect the conformance classes to know whether a particular feature is enabled or not.

It think the pgstac and sqlalchemy backends provide an implementation of the stac-fastapi framework. Perhaps the enabling of extensions could become a configurable option?

lossyrob:
The extensions added here are only for the example apps included in stac-fastapi. From what I can tell, normal usage would include a user creating their own instance of `StacApi` and configuring it with whatever extensions they would like (and therefore configurable in the way @rsmith013 requests). In this way, the addition of the ContextExtension here to the `app.py` examples in each of the backends that support it give the reader a better way to discover that extension.

@robintw how are you using stac-fastapi? Are you importing the `app` from one of the backend apps directly to get a FastAPI application? If so, you would have to rely on default configuration in these app.py's to change in order to change your configuration, though it seems like constructing your own instance of `StacApi` with all the configuration you need for your own instance of the server would give more flexibility.

robintw:
I originally started by importing `app` from one of the original configurations, but have since moved to constructing my own instance.

You're right that this would improve discoverability: I only found out about the Context Extension by finding that various other tools failed to interact with my STAC server properly. I did some Googling and found that they needed the Context Extension, and then a bit more searching showed me that it was implemented in stac-fastapi, but just not switched on by default. I agree that having it in the example apps would be useful.

(I might see if I can submit a PR with a bit of an update to the documentation emphasising that the defined `app` instances are just examples, and the best way to configure this for 'real-world' use is to define your own instance of `StacApi` with the configuration you want.)

lossyrob:
Thanks, that make sense - @jaysnm are you using the `app` in the same way as Robin was previous to creating an application-specific StacApi?

jaysnm:
Thanks for this conversation. Based on comments here, it turns out I have been running STAC back-end(s) the wrong way. In my team, we discourage building custom layer on top of upstream projects unless it's justifiably necessary (keeping track of updates and breaking changes adds to maintenance overheads). In practice thus, I directly use `app` from back-end to serve my STAC content. I will try this new approach of extending `app` to write my own custom StacAPI back-end. Probably this will also close issue #223. Yes I agree with @robintw, some of these examples can be pinned somewhere in the documentation. Once again, thank you all!

jaysnm:
> The extensions added here are only for the example apps included in stac-fastapi. From what I can tell, normal usage would include a user creating their own instance of `StacApi` and configuring it with whatever extensions they would lik 
>

This approach worked pretty fine for us and by so doing we were able to support other functionaries that were ambiguous previously. Our implementation was pretty simple!

```

"""STACnFastAPI application.
main.py
"""

from stac_fastapi.extensions.core import (
    FieldsExtension,
    QueryExtension,
    SortExtension,
    TransactionExtension,
    ContextExtension,
)
from stac_fastapi.extensions.third_party import BulkTransactionExtension
from stac_fastapi.sqlalchemy.config import SqlalchemySettings
from stac_fastapi.sqlalchemy.core import CoreCrudClient
from stac_fastapi.sqlalchemy.session import Session
from stac_fastapi.sqlalchemy.transactions import (
    BulkTransactionsClient,
    TransactionsClient,
)
from stac_fastapi.sqlalchemy.types.search import SQLAlchemySTACSearch

settings = SqlalchemySettings()
session = Session.create_from_settings(settings)
api = StacApi(
    settings=settings,
    extensions=[
        TransactionExtension(
            client=TransactionsClient(session=session), settings=settings
        ),
        BulkTransactionExtension(client=BulkTransactionsClient(session=session)),
        FieldsExtension(),
        QueryExtension(),
        SortExtension(),
        ContextExtension(),
    ],
    client=CoreCrudClient(session=session),
    search_request_model=SQLAlchemySTACSearch,
)

app = api.app

# add other stuff into app
# run the app using your favorite asgi server

```

> In this way, the addition of the ContextExtension here to the `app.py` examples in each of the backends that support it give the reader a better way to discover that extension.
>

Upstream repository is as good without this PR patches. Going forward, this PR can be closed. I'm happy to contribute into docs with an example `app` implemented from `StacApi` as a followup

moradology:
@jaysnm This needs a quick rebase and then I'd be happy to get it merged. Let me know if you can't find the time; I should be able to take care of things if that's helpful

jaysnm:
Hi @moradology. This is done and change-log updated. Let me know if any other change is need.

moradology (approved):

moradology:
Awesome. Thanks for the contribution!

jaysnm:
My pleasure.Thanks!

PR #213, opened 2021-08-10
refactoring to create dynamic request model

### LINKED ISSUE #203: Dynamic search request model based on extensions
At the moment, the `GET` and `POST` endpoints are based on static models and are different. 
Would be good to have a dynamic model which contains the core stac endpoints and is modified by the enabled extensions to represent the state of the API at runtime.

rsmith013:
closes #203 

I haven't made any changes to the sqlalchemy or pgstac modules. As I am not using either of them, I think it would be better for someone who is actively using them to try out the changes and make adjustments.

I have ended up with a Pydantic model for POST and an `APIRequest` for GET. This doesn't sit super comfortable but it seems necessary based on various threads on forums (e.g. [this one](https://bleepcoder.com/fastapi/456910758/question-using-pydantic-models-for-get-request-query-params)) which suggest that Pydantic is not appropriate for query params. 

## What's new?

This PR defines GET and POST request models for the extensions as well as pulling the pagination model out of the base. This should allow much more flexibility for implementation. Declared extensions are mixed together to create a dynamic request model which is rendered in the docs as query params for GET and body for POST.


lossyrob:
@rsmith013 what's the state of the PR? I'm curious if this is required with the recent move away from stac-pydantic, and also change in https://github.com/stac-utils/stac-fastapi/pull/271 to allow for custom search models to be used.

rsmith013:
@lossyrob I got as far as the changes required to the pgstac and sqlalchemy backends. As I don't use them, I wouldn't have a way to check that they were working properly. I can try and make the changes and then others can verify.

I'll take a deeper look at the other PR but this PR is about providing a framework where you don't need to define custom search models. The auto-generated docs reflect the implemented extensions.

This means that without any enabled extensions /search just provides basic stac parameters then implementing sort extension automatically adds the sort parameter, adding the filters extension automatically adds the filters parameter, etc...

It was about not baking in a particular implementation of the stac API and getting it to react to the enabled extension classes.

I will see if I can find time to update with master and push it to completion.

rsmith013:
@lossyrob This should be all done now. Let me know if you want anything changed. 
This hopefully moves to a standardised way of implementing extensions where it is declarative and the search request model is dynamically generated from the enabled extensions, including updating the conformance classes and creating a single, shared set of validators.

stac_fastapi/pgstac/stac_fastapi/pgstac/extensions/query.py
```diff
@@ -0,0 +1,48 @@
...
+from types import DynamicClassAttribute
+from typing import Any, Callable, Dict, Optional
+
+from pydantic import BaseModel
+from stac_pydantic.utils import AutoValueEnum
+
+from stac_fastapi.extensions.core.query import QueryExtension as QueryExtensionBase
+
+
+class Operator(str, AutoValueEnum):
```
moradology, line 14: I believe these operators find use in the `search` extension and not just `query`, so perhaps the previous spot (or a different file for operators) is more sensible than here

moradology (changes requested):
I've read through this PR a few times now - it everything looks consistent and clean. I'm happy to give this the +1 after a changelog entry is added.

One other note: I think we should be careful to bump the major version prior to releasing this code because it touches so much of the library

stac_fastapi/pgstac/stac_fastapi/pgstac/extensions/query.py
rsmith013: @moradology What is the search extension. Not come across that one. Happy to move it for the purposes of merging the PR.

rsmith013: As the query extension is deprecated in favour of filter extension and using something like [pygeofilter](https://github.com/geopython/pygeofilter) to parse the input for the filter extension will perform error checking I am not sure if this operator class has any use outside of the query extension. 

If not, I would say it should stay here so that it can be removed in a later release of stac-fastapi without worrying whether it is used by something else.

rsmith013:
@moradology I have added a changelog entry and merged master. I have held fire on moving the operator class for the reasons mentioned in the thread from your comment.

moradology (approved):

PR #214, opened 2021-08-10
Enable conformance class configuration

### LINKED ISSUE #169: API Landing Page not advertise conformance classes correctly
The landing page (root endpoint) does not have a conformsTo field, and the conformance classes in /conformance do not correspond to STAC API 1.0.0-beta.1 or 2. https://github.com/radiantearth/stac-api-spec/blob/master/overview.md#conformance-classes

This is an invalid STAC API as it works currently.

Planetary Computer does have a conformsTo field.

moradology:
This PR adds a `conformance_classes` parameter to the `LandingPageMixin` and all clients downstream from `BaseCoreClient` or `AsyncCoreClient`. This feature was added in a prior commit but appears to have been lost in the shuffle - a test has been added to hopefully avoid the regression going forward.

Closes #169

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
@@ -1,5 +1,5 @@
 """transaction extension."""
-from typing import Callable, Type, Union
+from typing import Callable, List, Type, Union
 
 import attr
 from fastapi import APIRouter, FastAPI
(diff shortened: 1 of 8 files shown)


philvarner:
I just ran this against the STAC API Validator and is passed the conformance part of the tests (though it blows up because the /collections endpoint is wrong)

geospatial-jeff (approved):
LGTM! There will be some merge conflicts due to #199

rsmith013:
@moradology In the filter extension PR, I added a method `list_conformance_classes` to generate the list of conformance classes based on which extensions were enabled by the server. I think it would be good to move this PR in that direction rather than adding an extra instance attribute to the core client which needs to be configured. Thoughts?

rsmith013:
nice!

lossyrob:
What is the merge strategy here with #165? Should this be updated and merged, and then #165 updated with these changes, or do the changes not touch each other?

lossyrob:
Sounds good - @moradology can you update to latest and then I'll merge?

moradology:
@lossyrob Pinging you so that you can push the button. Should be good to go

Issue #220, opened 2021-08-11
/search endpoint returns Content-Type header of application/json instead of application/geo+json


philvarner:
/search endpoint returns Content-Type header of application/json instead of application/geo+json as required by the STAC API spec.

This was tested with a local docker-compose instance of stac-fastapi-sqlalchemy.

Issue #219, opened 2021-08-11
Landing page `search` link relation advertised with type application/json instead of application/geo+json


philvarner:
Landing page `search` link relation advertised with type application/json instead of application/geo+json

This was tested with a local docker-compose instance of stac-fastapi-sqlalchemy.

Issue #218, opened 2021-08-11
Incorrect "conformance" link relation href in Landing Page


philvarner:
The Landing Page contains a link relation for `conformance` relation, but this links to the landing page rather than `/conformance`.

This was tested with a local docker-compose instance of stac-fastapi-sqlalchemy.

duckontheweb:
@philvarner I'm not able to replicate this on `main`, it appears the `conformance` link correctly points to `/conformance`. Here are the steps I took:

```console
$ docker-compose up --build app-sqlalchemy database loadjoplin-sqlalchemy
$ curl http://localhost:8081/ | jq '.links[] | select(.rel == "conformance")'
{
  "rel": "conformance",
  "type": "application/json",
  "title": "STAC/WFS3 conformance classes implemented by this server",
  "href": "http://localhost:8081/conformance"
}
```

Do you maybe need to rebuild the images?

philvarner:
I was rebuilding, but I may have been running from someone else's branch that was out of date. Looks fixed now.

philvarner:
CNR

Issue #217, opened 2021-08-11
Missing service-desc and service-doc link relations in Landing Page


philvarner:
The landing page should contain these two endpoints, but does not:
- / : Link[rel=service-desc] (required by all conformance classes)
- / : Link[rel=service-doc] (optional, but recommended)

This was tested with a local docker-compose instance of stac-fastapi-sqlalchemy.

philvarner:
cnr

PR #229, opened 2021-08-12
Fix/add landing page links

duckontheweb:
**Related Issues:**

- #219 
- #218 
- #217 
- #216 

**Summary of Changes:**

- Uses a module-scoped fixture to fetch the landing page within `test_conformance` so that we only fetch it once
- Adds parameterized tests for links related to the above issues
- Adds `root` and `service-desc` links to landing page
- Changes `type` of `search` link to `"application/geo+json"`

**Questions:**

- [x] Should the `"service-desc"` rel type be added to `stac-pydantic.links`?
- [x] Should the `"application/vnd.oai.openapi+json;version=3.0"` MIME type be added to `stac_pydantic.shared`?
   Opened stac-utils/stac-pydantic#101 for both of these changes.
- [x] Should we [change the default OpenAPI endpoint](https://fastapi.tiangolo.com/advanced/conditional-openapi/?h=openapi#conditional-openapi-from-settings-and-env-vars) to be `/api` to be in line with the [recommendation in the STAC API spec](https://github.com/radiantearth/stac-api-spec/tree/master/core#link-relations)?
   This is configurable in `ApiSettings.openapi_url` and defaults to `/api`
- [ ] ~**UPDATE:** Should I also include a fix to the `/search` endpoint `"Content-Type"` header for #220?~ I'll do this in a separate PR since it seems like it might be a bit involved.

~I plan to add the equivalent changes to the `sqlalchemy` app but was having some trouble getting those tests to run and I thought I'd get feedback on this first.~ I got these working and added the equivalent tests.

cc: @philvarner

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,10 @@ run-database:
 run-joplin-sqlalchemy:
 	docker-compose run --rm loadjoplin-sqlalchemy
 
+.PHONY: run-joplin-sqlalchemy
+run-joplin-sqlalchemy:
+	docker-compose run --rm loadjoplin-sqlalchemy
+
 .PHONY: test
 test: test-sqlalchemy test-pgstac
 
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -62,7 +62,11 @@ class StacApi:
     exceptions: Dict[Type[Exception], int] = attr.ib(
         default=attr.Factory(lambda: DEFAULT_STATUS_CODES)
     )
-    app: FastAPI = attr.ib(default=attr.Factory(FastAPI))
+    app: FastAPI = attr.ib(
+        default=attr.Factory(
+            lambda self: FastAPI(openapi_url=self.settings.openapi_url), takes_self=True
+        )
+    )
     router: APIRouter = attr.ib(default=attr.Factory(APIRouter))
     title: str = attr.ib(default="stac-fastapi")
     api_version: str = attr.ib(default="0.1")
--- a/stac_fastapi/pgstac/tests/clients/test_postgres.py
+++ b/stac_fastapi/pgstac/tests/clients/test_postgres.py
@@ -80,7 +80,7 @@ async def test_update_item(app_client, load_test_collection, load_test_item):
 
     item.properties.description = "Update Test"
 
-    resp = await app_client.put(f"/collections/{coll.id}/items", data=item.json())
+    resp = await app_client.put(f"/collections/{coll.id}/items", content=item.json())
     assert resp.status_code == 200
 
     resp = await app_client.get(f"/collections/{coll.id}/items/{item.id}")
@@ -113,7 +113,7 @@ async def test_get_collection_items(app_client, load_test_collection, load_test_
         item.id = str(uuid.uuid4())
         resp = await app_client.post(
             f"/collections/{coll.id}/items",
-            data=item.json(),
+            content=item.json(),
         )
         assert resp.status_code == 200
 
--- a/stac_fastapi/pgstac/tests/conftest.py
+++ b/stac_fastapi/pgstac/tests/conftest.py
@@ -111,7 +111,7 @@ async def app(api_client):
 
 
 @pytest.mark.asyncio
-@pytest.fixture()
+@pytest.fixture(scope="session")
 async def app_client(app):
     async with AsyncClient(app=app, base_url="http://test") as c:
         yield c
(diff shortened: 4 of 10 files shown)


lossyrob:
Looks like this has some changes from #230 , we should merge that first and then update this branch.

duckontheweb:
> Looks like this has some changes from #230 , we should merge that first and then update this branch.

I cherry picked the makefile change to make the sqlalchemy tests run, but other than that these should be able to be merged independently.

philvarner:
> Should we change the default OpenAPI endpoint to be /api to be in line with the recommendation in the STAC API spec?

In some way, I would wish that you didn't change it, so that a prominent implementation would be doing something other than /api, which would allow us to more easily find clients who hardcode /api rather than discovering the URL from the root links.

geospatial-jeff:
It might be worth removing the openapi / docs endpoints because I don't think they will work with changes in #147

duckontheweb:
> It might be worth removing the openapi / docs endpoints from the landing page because I don't think they will work with changes introduced in #147

They seem to be working okay on `master` right now. I'm serving up the `pgstac` dev app from the current `master` and the `docs`, `redoc`, and `openapi.json` endpoints are giving (what I think) are reasonable responses. Is there something in particular that might be broken about those?

duckontheweb:
> I would recommend against removing them, even if they're not entirely accurate, as the link relation is required by stac api.

@philvarner To me the STAC API docs are a little fuzzy on this point. There's this sentence from the [Core API docs](https://github.com/radiantearth/stac-api-spec/blob/8745df777e33dbd6c9a4eb684568d94825940f09/core/README.md)...

> This spec does not require any API endpoints from OAFeat or STAC API to be implemented, so the following links may not exist if the endpoint has not been implemented.

...and then in the [Link Relations](https://github.com/radiantearth/stac-api-spec/blob/master/core/README.md#link-relations) section (emphasis added)...

> The following Link relations *should* exist in the Landing Page (root).

However, in reality, it seems like `root`, `self`, and `service-desc` *must* be present and `child` *may* be present. Maybe we should reorganize those links a bit for clarity or change the wording?

That aside, I agree we should keep the links even if they're not perfect, as long as they are not misleading.

duckontheweb:
> > Should we change the default OpenAPI endpoint to be /api to be in line with the recommendation in the STAC API spec?
> 
> In some way, I would wish that you didn't change it, so that a prominent implementation would be doing something other than /api, which would allow us to more easily find clients who hardcode /api rather than discovering the URL from the root links.

I can see the usefulness of this from the client validation standpoint and I don't have a strong opinion on this, so I will go with whatever the consensus is. The thinking for changing it is that it seems like `stac-fastapi` is quickly becoming the de facto Python STAC API implementation, so it seems like it should follow best practices as closely as possibly.

philvarner:
> > I would recommend against removing them, even if they're not entirely accurate, as the link relation is required by stac api.
> 
> @philvarner To me the STAC API docs are a little fuzzy on this point. There's this sentence from the [Core API docs](https://github.com/radiantearth/stac-api-spec/blob/8745df777e33dbd6c9a4eb684568d94825940f09/core/README.md)...

I agree that's not worded well. For one, that should be "shall" and not "should". (I shall fix it, if you should like)

Moving `child` out is a good idea.

philvarner:
> I can see the usefulness of this from the client validation standpoint and I don't have a strong opinion on this, so I will go with whatever the consensus is. The thinking for changing it is that it seems like `stac-fastapi` is quickly becoming the de facto Python STAC API implementation, so it seems like it should follow best practices as closely as possibly.

Definitely, stac-fastapi is that. Good point about following best practices

duckontheweb:
> > > I would recommend against removing them, even if they're not entirely accurate, as the link relation is required by stac api.
> > 
> > 
> > @philvarner To me the STAC API docs are a little fuzzy on this point. There's this sentence from the [Core API docs](https://github.com/radiantearth/stac-api-spec/blob/8745df777e33dbd6c9a4eb684568d94825940f09/core/README.md)...
> 
> I agree that's not worded well. For one, that should be "shall" and not "should". (I shall fix it, if you should like)
> 
> Moving `child` out is a good idea.

😆  I should!

duckontheweb:
> > I can see the usefulness of this from the client validation standpoint and I don't have a strong opinion on this, so I will go with whatever the consensus is. The thinking for changing it is that it seems like `stac-fastapi` is quickly becoming the de facto Python STAC API implementation, so it seems like it should follow best practices as closely as possibly.
> 
> Definitely, stac-fastapi is that. Good point about following best practices

The OpenAPI endpoint is now configurable with `ApiSettings.openapi_url` and defaults to `/api`.

duckontheweb:
@geospatial-jeff @lossyrob This is ready for another look-over

lossyrob (approved):

PR #230, opened 2021-08-12
Fix test-* make targets

duckontheweb:
**Changes:**
- Adds a `run-database` make target to run just the `database` service for testing
   This was purely for convenience

- Uses `app-pgstac` instead of `app-sqlalchemy` in `test-pgstac` target
- Removes `test-pgstac` dependency on `pgstac-install` target
   This was causing problems when installing `rasterio` in my environment and we don't appear to need it since we are running the tests inside the `stac-fastapi-pgstac` container.
- **UPDATE:** Adds `run-joplin-sqlalchemy` make target and adds it as a dependency to `test-sqlalchemy` to ensure test data is loaded prior to running unit tests.

Again, could be some user error here, so let me know if I'm off-base.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,12 @@ run_docker = docker-compose run --rm \
 				-e APP_PORT=${APP_PORT} \
 				app-sqlalchemy
 
+run_pgstac = docker-compose run --rm \
+				-p ${EXTERNAL_APP_PORT}:${APP_PORT} \
+				-e APP_HOST=${APP_HOST} \
+				-e APP_PORT=${APP_PORT} \
+				app-pgstac
+
 .PHONY: image
 image:
 	docker-compose build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,12 +27,20 @@ docker-shell:
...
 
 .PHONY: test-sqlalchemy
-test-sqlalchemy:
+test-sqlalchemy: run-joplin-sqlalchemy
 	$(run_docker) /bin/bash -c 'export && cd /app/stac_fastapi/sqlalchemy/tests/ && pytest'
 
 .PHONY: test-pgstac
-test-pgstac: pgstac-install
-	$(run_docker) /bin/bash -c 'export && cd /app/stac_fastapi/pgstac/tests/ && pytest'
+test-pgstac:
```
lossyrob, line 34: Should this also have a dependency on a `run-joplin-pgstac` target?

duckontheweb: I cleared my Docker caches using:

```
docker system prune
docker volume prune
```
and then ran the tests using:
```
make run-database
make test-pgstac
```
and all of the tests succeeded, which made me think that it wasn't relying on that data. We could add that as a dependency just in case we write tests later that need it.

lossyrob: Interesting - seems like the pgstac test were written to not rely on the dev datasets, which I think is a better approach. Thanks for clearing that up.

lossyrob (approved):

Issue #172, opened 2021-07-03
pgstac: Return empty item collection instead of error for empty search


lossyrob:
[Here](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py#L180) pgstac returns a 404 if no items are found. I think a more expected response would be a 200 with an empty ItemCollection, as the search endpoint is found and is working, and the empty search results is a valid result.

zstatmanweil:
@lossyrob: I was about to open an issue for this but found this old one. Looks like you made the change at some point but that branch never made it into a PR. Do you think I can/should open a PR that just isolates the fix for this issue? I don't have much experience contributing to open source so trying to figure out what is appropriate!

moradology:
Yeah, this is one that I could have sworn we'd taken care of but appears to have slipped through the cracks. I think this would be a great contribution and I'd be happy to pick it up myself but if you'd like to take a shot @zstatmanweil, I'd be more than happy to lend a hand when/where useful

zstatmanweil:
Thanks to you both! @moradology, I will give it a shot and let you know if I have questions. I will just implement the change relating to this issue though. I will tag you in the PR when I open it

zstatmanweil:
Thanks for your help with my first PR @lossyrob, @moradology. Looking forward to contributing more! 

We can close this issue now, ya?

moradology:
Absolutely and no problem at all. And in the future, you can replace "This PR addresses issue #172." with "closes #172" and merging the PR will automatically close the referenced issue

PR #233, opened 2021-08-12
Return empty item collection instead of error when searching

zstatmanweil:
This PR addresses issue https://github.com/stac-utils/stac-fastapi/issues/172. 
- A POST to `/search` that returns no items now returns an empty list of features rather than a 404 and `NotFoundError`;
- I kept the logic that a GET `/collections/{collectionId}/items` with an invalid collection ID still returns a 404 rather than a 200 with an empty features list 
- I added logic such that a GET to `/collections/{collectionId}/items/{itemId}` with an invalid collection ID returns a 404

Testing:
- Added a few unit tests and all unit tests are passing
- Ran locally and tested behavior as expected

cc: @moradology, @lossyrob


stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -161,13 +159,16 @@ async def item_collection(
         Returns:
             An ItemCollection.
         """
+        # If collection does not exist, NotFoundError wil be raised
```
moradology, line 162: minor typo here

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -161,13 +159,16 @@ async def item_collection(
...
+        await self.get_collection(id, **kwargs)
```
moradology, line 163: I'm curious what convinced you to safeguard against non-existent collections here. Maybe you saw some performance hits in that case which can be avoided?

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -180,9 +181,15 @@ async def get_item(self, item_id: str, collection_id: str, **kwargs) -> Item:
         Returns:
             Item.
         """
+        # If collection does not exist, NotFoundError wil be raised
+        await self.get_collection(collection_id, **kwargs)
```
moradology, line 185: Same question as above regarding the benefits of this safeguard

moradology (approved):
I found an incredibly minor issue and one bit that I had some questions about, but this PR looks good to me. Nice tests in particular.

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
zstatmanweil: This addition is to maintain the existing logic that a request to this endpoint would result in an error if a collection doesn't exist. Previously it was the `item_collection = await self._search_base(req, **kwargs)` that would result in a 404 since no items would be found, but since that now returns an empty collection and 200 if no items are found, I wanted to make sure an error was raised. 

I was most unsure about this change, especially since The Planetary Computer STAC actually doesn't raise an error in this instance when a collection doesn't exist. I think logically it should though. What do you think?

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
zstatmanweil: Same idea as above - checking if a collection exists because if not a 404 will be returned

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
lossyrob: Good catch - the PC should def return a 404 instead of an empty collection in this case, so this change LGTM

geospatial-jeff (approved):

PR #238, opened 2021-08-18
Upgrade to pgstac 0.3.2

lossyrob:
This PR updates pgstac to 0.3.2

It also adds a few convenience Makefile commands for pgstac, accounts for https://github.com/stac-utils/pgstac/issues/45 by guarding against a null case, and fixes a tab-typo that was causing a list to be set in a loop - which only started failing tests when the collection['features'] was null.


stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -129,7 +129,7 @@ async def _search_base(
         collection = ItemCollection(**items)
         cleaned_features: List[Item] = []
 
-        for feature in collection["features"]:
+        for feature in collection.get("features") or []:
```
kylebarron, line 132: you can't do `.get('features', [])`? Most `.get` operators in Python allow for a second fallback argument, which just defaults to None

lossyrob: This doesn't just account for the case where there is no `features` key, but also where there is a `features` key and the value is `None`

moradology (approved):

PR #240, opened 2021-08-19
Update pgstac to return 400 on invalid date parameter.

lossyrob:
pgstac currently returns a 500 with an invalid date; this PR modifies this to return a 400.

Also allow backends to raise their own InvalidQueryParameter exceptions, which will be translated into a 400 by exception handling. This is another option for backends that want to handle query parameter validation in other ways besides stac-pydantic model validation.

[This line](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py#L109) in the pgstac search model indicates that validation is done outside of the pydantic model, but there were instances of `asyncpg.exceptions.InvalidDatetimeFormatError` being thrown for bad datetime parameters. We could solve this by removing the overridden validator, but I'm not sure why this was overridden and also I think it's useful to provide a non-pydantic mechanism for backends to indicate parameters are invalid and having stac-fastapi return the right status code.

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/errors.py
+++ b/stac_fastapi/api/stac_fastapi/api/errors.py
@@ -13,6 +13,7 @@ from stac_fastapi.types.errors import (
     ConflictError,
     DatabaseError,
     ForeignKeyError,
+    InvalidQueryParameter,
     NotFoundError,
 )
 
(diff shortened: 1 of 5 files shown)


moradology (approved):
This looks good to me. Removing a print statement that I assume was added in error and merging in master. Once CI passes, I'll merge

PR #270, opened 2021-09-23
Set collection_id on search for pgstac get_item; update to pgstac 0.3.4

lossyrob:
**Description:**
This change modifies the behavior of get_item in the pgstac backend to set the collection ID. This is correct behavior - item Ids must be unique within a collection, and a collection_id is supplied in the call. Also, as mentioned in https://github.com/stac-utils/pgstac/issues/58, pgstac can return incorrect results if queried with just an Item ID without the Collection ID set. 

This PR also updates to pgstac 0.3.4, and fixes the NotFoundError message in get_item.

**PR Checklist:**


moradology:
Nice to get this in before cutting the next release. Everything seems to be doing what it's supposed to, so 👍

moradology (approved):

PR #271, opened 2021-09-27
Allow custom models for search GET and items endpoints

lossyrob:
**Description:**
I use a different default limit than this project, which is 10. While currently you can override the `search_request_model` in the StacApi class, you can't do so for the models that give default values to the item collection endpoint and the GET search. This PR adds models that allow users to override these types. It also contains the ability for the pgstac backend to use a custom search model instead of building a PgstacSearch for the item collection and GET search endpoints. I made this change while tracking down why my overridden limit was not being used; this turns out not being needed for my specific use case (as the default limit comes from the arguments passed into the core client methods by ItemCollectionUri and SearchGetRequest), it seemed like a sensible change in case any other user would want to modify the behavior of the search model in more than just the search POST endpoint.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
@@ -1,7 +1,7 @@
 """Item crud client."""
 import re
 from datetime import datetime
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Type, Union
 from urllib.parse import urljoin
 
 import attr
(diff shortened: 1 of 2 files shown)


stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -27,6 +27,8 @@
 class CoreCrudClient(AsyncBaseCoreClient):
     """Client for core endpoints defined by stac."""
 
+    search_request_model: Type[PgstacSearch] = attr.ib(init=False, default=PgstacSearch)
```
moradology, line 30: It might be possible to avoid this duplication of `search_request_model` specification if the client methods that use it were higher order functions. This would mean taking the `StacApi` specified `search_request_model` and returning a function which would then be registered as an endpoint.

moradology (approved):
This looks good. There's some duplication of arguments but the fix (described in a below comment) might not be desirable. +1 if the duplication fix isn't something we want to pursue

PR #282, opened 2021-11-01
Return existing links in pgstac

lossyrob:
**Description:**
Collections and Items stored in pgstac may have existing links. The example that uncovered this bug is for Collections to link to their licenses with rel=license link. These are currently being overwritten by the hierarchical links derived by stac-fastapi. This change passes through any non-hierarchical links in the pgstac backend.

This also changes the joplin ingest code to use a PUT if the collection or item already exists - useful for changing the development data and re-ingesting.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
@@ -23,18 +23,6 @@ from stac_fastapi.types.stac import Collection, Collections, Item, ItemCollectio
 NumType = Union[float, int]
 
 
-def _extended_links(
-    original_links: Optional[List[Dict[str, Any]]], new_links: List[Dict[str, Any]]
-) -> List[Dict[str, Any]]:
-    """Extend links with new links."""
-    if original_links is None:
-        return new_links
-    new_link_rels = set([link["rel"] for link in new_links])
-    return new_links + [
-        link for link in original_links if link["rel"] not in new_link_rels
-    ]
-
-
 @attr.s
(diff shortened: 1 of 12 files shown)


stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -23,6 +23,18 @@
...
+) -> List[Dict[str, Any]]:
+    """Extend links with new links."""
+    if original_links is None:
+        return new_links
+    new_link_rels = set([link["rel"] for link in new_links])
+    return new_links + [
+        link for link in original_links if link["rel"] not in new_link_rels
+    ]
+
+
```
moradology, line 37: I'm wondering if this shouldn't exist in the `get_links` method under logic triggered by the `extra_links` argument, here: https://github.com/stac-utils/stac-fastapi/blob/b26a12ae210ac65fed2fe9344a13c1357281dce5/stac_fastapi/pgstac/stac_fastapi/pgstac/models/links.py#L94-L98

I believe the intent of that argument is to allow extensions of the generated links; it appears to not be used in the updated portions here

moradology (changes requested):
It looks to me as though the existing machinery for link extension should be used and updated if necessary rather than the creation of an `_extended_links` method

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
lossyrob: Good call, changed in 4c37355

moradology (approved):

Issue #281, opened 2021-10-29
'/api' route does not handle 'Accept' header for OGC API Feautes - part 2 spec requirement. 


NicolaiMogensen:
'/api' route does not handle 'Accept' header for OGC API Feautes - part 2 spec requirement. Content-type of the route `/api` should be `content-type: application/vnd.oai.openapi+json;version=3.0`. Probably all that needs here is to instead of using JSONResponse, use something like:

```python
class VndResponse(JSONResponse):
    media_type = application/vnd.oai.openapi+json;version=3.0
```

Recreate problem using teamengine:
http://cite.opengeospatial.org/teamengine/ ( or for local instances, a docker image is supplied here: https://cite.opengeospatial.org/teamengine/about/ogcapi-features-1.0/1.0/site/ see the docker section log in with ogctest:ogctest. If using docker you have to use the ip for localhost, use ipconfig or similar to find it)

Teamengine will check the OGC API implementation against spec, and notify you of any problems. Since Stac-Api is based on OGC API i would assume it to pass this check. 

The teamengine test fails this task: 

<html>
<body>

Test Name: | api Definition Validation
-- | --
Test Description: | Implements A.2.3. API Definition Path {root}/api (link), Abstract Test 6 (Requirement /req/core/api-definition-success) \| Details ↗Class Name:ApiDefinitionMethod Name:apiDefinitionValidationPath:org/opengis/cite/ogcapifeatures10/conformance/core/apidefinition/ApiDefinition
Test Result: | FAILED

</body>
</html>

In laymans terms this test refers to https://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_api_definition_path_rootapi_link which states that api-definition-success (https://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_response_2)

should fulfill requirement 4 which states:

<html>
<body>

Requirement 4 | /req/core/api-definition-success
-- | --
A | A GET request to the URI of an API definition linked from the landing page (link relations service-desc or service-doc) with an Accept header with the value of the link property type SHALL return a document consistent with the requested media type.

</body>
</html>

And right now the requested media-type is supposed to be `content-type: application/vnd.oai.openapi+json;version=3.0` but is simply `application/json` and so the test fails. 

I have tried looking into where to implement this, but i fail to see how exactly the `/api` route is handled, maybe it's something core to `FastAPI` itself.

PR #287, opened 2021-11-08
Fix content-type of /api response

moradology:
**Related Issue(s):** #281

**Description:**
/api provides open api JSON which is used for doc generation. This endpoint currently returns `application/json` as the content type, whereas OGC API Features - part 2 spec explicitly requires `application/vnd.oai.openapi+json;version=3.0` to be the content-type. This PR addresses that issue by extending the `FastAPI` class to update the `setup` method, which is hides the creation of the /api endpoint.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -4,13 +4,20 @@ from typing import Any, Callable, Dict, List, Optional, Type, Union
 import attr
 from brotli_asgi import BrotliMiddleware
 from fastapi import APIRouter, FastAPI
+from fastapi.openapi.docs import (
+    get_redoc_html,
+    get_swagger_ui_html,
+    get_swagger_ui_oauth2_redirect_html,
+)
 from fastapi.openapi.utils import get_openapi
+from fastapi.responses import ORJSONResponse
 from pydantic import BaseModel
 from stac_pydantic import Collection, Item, ItemCollection
 from stac_pydantic.api import ConformanceClasses, LandingPage, Search
 from stac_pydantic.api.collections import Collections
 from stac_pydantic.version import STAC_VERSION
-from starlette.responses import JSONResponse, Response
+from starlette.requests import Request
+from starlette.responses import HTMLResponse, JSONResponse, Response
 
 from stac_fastapi.api.errors import DEFAULT_STATUS_CODES, add_exception_handlers
 from stac_fastapi.api.models import (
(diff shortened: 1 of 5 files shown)


stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -32,6 +39,69 @@
...
+class VndOaiResponse(ORJSONResponse):
+    """JSON with custom, vendor content-type."""
+
+    media_type = "application/vnd.oai.openapi+json;version=3.0"
+
+
+class VndFastAPI(FastAPI):
+    """Custom FastAPI to support /api response with vendor content-type."""
+
+    def setup(self) -> None:
```
lossyrob, line 51: Perhaps unnecessary code duplication can be avoided by calling `super` and then removing/re-adding the route itself.

`self` is a `Starlette` object, which has a [routes](https://github.com/encode/starlette/blob/master/starlette/applications.py#L95) property that directly returns the list of the router's routes. AFAIK adding a route appends it to the list and does no other special registration. While there's no "remove_route" method, if you find the route with the `self.openapi_url` and replace it with the one you construct with the proper response, that might save some code dupe. The worry with the dupe is that if a later version of FastAPI changes say the docs_url add logic, then this override logic won't pick up on those changes unrelated to the VndOaiResponse.

Another thing to note is that [Route](https://github.com/encode/starlette/blob/6af5c515e0a896cbf3f86ee043b88f6c24200bcf/starlette/routing.py#L189) exposes the endpoint Callable directly, so you could even wrap the endpoint in one that translates the return value into a VndOaiResponse.

Maybe not worth spending too much time trying to hack the internals of FastAPI and Starlette (though everything I mentioned is public API), but I think it would be preferable to reduce the code copy in the setup override to only as much as is needed for this small change.

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -75,7 +146,7 @@ class StacApi:
     search_request_model: Type[Search] = attr.ib(default=STACSearch)
     search_get_request: Type[SearchGetRequest] = attr.ib(default=SearchGetRequest)
     item_collection_uri: Type[ItemCollectionUri] = attr.ib(default=ItemCollectionUri)
-    response_class: Type[Response] = attr.ib(default=JSONResponse)
+    response_class: Type[Response] = attr.ib(default=ORJSONResponse)
```
lossyrob, line 149: This changes the default JSONResponse value to ORJSONResponse, which seems like an orthogonal change to this PR. And, orjson isn't an explicit dependency to anything but pgstac, so either orjson should be added as a dependency to the `api` project or it shouldn't be used in this code file (applies also to the base class of `VndOaiResponse), as it would break installations that only relied on the sqlalchemy backend.

lossyrob (changes requested):

stac_fastapi/api/stac_fastapi/api/app.py
moradology: It appears as though `attrs` has a `converter` argument, which allows you to make modifications to supplied arguments. I've implemented a function which can be used as a `converter` to alter the routes such that the appropriate vendor content-type is advertised without any inheritance shenanigans

lossyrob (approved):

PR #288, opened 2021-11-09
Set content-type to geojson for search results

moradology:
**Related Issue(s):** #220

**Description:**
This PR changes the /search endpoint's headers to reflect the fact that it returns geojson rather than mere json, which is in line with the STAC api spec

**PR Checklist:**

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -97,7 +97,7 @@ class StacApi:
         self,
         func: Callable,
         request_type: Union[Type[APIRequest], Type[BaseModel]],
-        resp_class: Type[Response] = response_class,
+        resp_class: Type[Response],
     ) -> Callable:
         """Create a FastAPI endpoint."""
         if isinstance(self.client, AsyncBaseCoreClient):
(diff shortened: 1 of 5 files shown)


stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -175,12 +181,12 @@ def register_post_search(self):
             response_model=(ItemCollection if not fields_ext else None)
             if self.settings.enable_response_models
             else None,
-            response_class=self.response_class,
+            response_class=GeoJSONResponse,
```
lossyrob, line 184: This will cause the GeoJSONResponse to not use ORJSONResponse if the `self.response_class` is ORJSONResponse, leading to slowdowns in what's the largest API responses.

Is there a way to just set the `media_type` at response class creation time instead of setting the media type through this mechanism?

lossyrob (changes requested):

stac_fastapi/api/stac_fastapi/api/app.py
moradology: To resolve this, I've used `importlib` to conditionally use `ORJSONResponse` when `orjson` is available and `JSONResponse` otherwise.

Issue #290, opened 2021-11-09
500s thrown for large limit searches


moradology:
- GET Search with {'limit': 10000} returned status code 500
- POST Search with {'limit': 10000} returned status code 500

moradology:
This appears to be implementation specific and can't be considered an issue with stac-fastapi.  To control the possible size of `limit` arguments during search, see the discussion of `STACSearch` inheritance in #297

PR #297, opened 2021-11-16
Allow users configuration of search limit

moradology:
**Related Issue(s):** #290

**Description:**
This PR enables end users to customize the maximum `limit` value for search parameters. The spec *suggests* 10000 as the maximum limit but implementations are free to set their own. This fix should allow implementers to put a limit on extremely large requests as may be necessary to avoid errors (the source of the 500s mentioned in issue #290) or to limit costs

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+* Users can now specify a maximum `limit` value through the `max_search_limit` configuration value ([290](https://github.com/stac-utils/stac-fastapi/pull/290))
+
 ### Changed
 
 ### Removed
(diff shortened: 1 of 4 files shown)


geospatial-jeff:
@moradology A user can already set their own `limit` parameter by overriding the pydantic model used for searches (https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/stac_fastapi/api/app.py#L75-L76).  I'm not sure its a good idea to have multiple ways to override search parameters, but I don't really have an opinion on which one is preferred.  What do you think?

moradology:
@geospatial-jeff I wasn't initially in agreement with your comment but after reflecting on it for a bit, I think you're right that handling this by inheritance is the cleanest solution. I'm going to close this PR; I am only left wondering if the change is obvious enough for users to reliably figure out where this particular 400 error is coming from

Issue #284, opened 2021-11-04
Generalize the item geometry type on the Sqlalchemy backend


yellowcap:
## Problem
I ran into some trouble when ingesting STAC items that do not have a simple Polygon geometry through the sqlalchemy backend. 

The geometry field is [required to be a Polygon](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/models/database.py#L67), although the STAC Item spec specifies the field as the more general [GeoJSON Geometry](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields), which includes other types such as MultiPolygon.

## Example STAC Item with MultiPolygon
I came across this problem when trying to ingest STAC Item data from the [usgs-landsat public bucket](https://registry.opendata.aws/usgs-landsat/).

In this catalog, there are some items that have the MultiPolygon type. For instance, Landsat scenes at the edge of the WGS84 projection are split into two. Here is a link to an example:

s3://usgs-landsat/collection02/level-2/standard/etm/2021/092/013/LE07_L2SP_092013_20211007_20211104_02_T2/LE07_L2SP_092013_20211007_20211104_02_T2_SR_stac.json

## Proposed solution
I would like to propose to change the geometry field to MultiPolygon or even GeometryCollection. This would still be matching the STAC spec.

This would presumably require a change in the field definition, an additional migration, and maybe a data-migration if the field content needs to be altered after this change. For instance, converting existing Polygon data to MultiPolygon.

## Related
The change of the field type was previously proposed, but the PR was mixed with a few other changes that were unrelated, so I am opening the discussion again. See https://github.com/stac-utils/stac-fastapi/pull/184

geospatial-jeff:
I agree that the geometry type used by sqlalchemy backend needs to better align with the spec.  `GeometryCollection` type would probably be best because it gives the most flexibility.  My one concern is I'm not sure if there are performance drawbacks to using a `GeometryCollection` type over a more explicit type like `MultiPolygon` or `Polygon` (currently); that might be worth exploring before we make a change.

For what its worth, stac-pydantic recently added support for specifying `GeometryCollection` in search requests (https://github.com/stac-utils/stac-pydantic/pull/102), this change will be shipped in the next stac-pydantic release.

photonbit:
[...]

PR #299, opened 2021-11-23
Change geometry type in the items table

yellowcap:
**Related Issue(s):** 
https://github.com/stac-utils/stac-fastapi/issues/284

**Description:**
This generalizes the geometry type of the geometry column in the Items table from `Polygon` to `Geometry`.

- As far as I understood, `Polygon` is a subset of `Geometry`, so this migration does not require a data migration step.
- I am not sure about the downgrade migration. It will only work if no other geometries are added to the DB after the change. So maybe there the alternative is simply be a `pass`, as the data in the table might not be backwards compatible. On the other hand, the way its proposed its the proper downgrade from a type perspective.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/sqlalchemy/alembic/versions/5909bd10f2e6_change_item_geometry_column_type.py
+++ b/stac_fastapi/sqlalchemy/alembic/versions/5909bd10f2e6_change_item_geometry_column_type.py
@@ -1,0 +1,34 @@
+"""change item geometry column type
+
+Revision ID: 5909bd10f2e6
+Revises: 821aa04011e8
+Create Date: 2021-11-23 10:14:17.974565
+
+"""
+from alembic import op
+
+from stac_fastapi.sqlalchemy.models.database import GeojsonGeometry
+
+# revision identifiers, used by Alembic.
+revision = "5909bd10f2e6"
+down_revision = "821aa04011e8"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column(
+        schema="data",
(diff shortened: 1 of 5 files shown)


stac_fastapi/sqlalchemy/alembic/versions/5909bd10f2e6_change_item_geometry_column_type.py
```diff
@@ -0,0 +1,35 @@
+"""change item geometry column type
+
+Revision ID: 5909bd10f2e6
+Revises: 821aa04011e8
+Create Date: 2021-11-23 10:14:17.974565
+
+"""
+import sqlalchemy as sa
```
lossyrob, line 8: Checked linting, just need to remove this line for flake8 to pass

yellowcap: Thanks @lossyrob - just did that and also rebased from current master to avoid conflicts.

yellowcap:
I added a copy of a stac item from the L28 L2 collection that has a multipolygon feature (the one that I referenced in the issue https://github.com/stac-utils/stac-fastapi/issues/284). 

With that I added a test for ingesting the item and running a search against it. I am not sure if that is sufficient to properly test the change. So @lossyrob , when you have a moment please have a look and let me know if anything is missing.

lossyrob (approved):

yellowcap:
:+1: thanks for your help on this @lossyrob

Issue #302, opened 2021-11-29
Deleting, Updating items should only work with specifying a collection - SQLAlchemy


jonhealy1:
Different collections can have items with the same ids. To delete or update an item, a collection should be specified. In core.py for sqlalchemy, get_item was recently changed so that only an item from the right collection is returned. This is a similar idea.

PR #303, opened 2021-11-29
Update item transactions in SQLAlchemy

jonhealy1:
**Related Issue(s):** #302 

**Description:**

Different collections can have items with the same ids. To delete or update an item, a collection should be specified. In core.py for sqlalchemy, get_item was recently changed so that only an item from the right collection is returned. This is a similar idea.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
+++ b/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
@@ -62,7 +62,9 @@ class TransactionsClient(BaseTransactionsClient):
                 self.item_table.collection_id == model["collection"]
             )
             if not query.scalar():
-                raise NotFoundError(f"Item {model['id']} in collection {model['collection']}")
+                raise NotFoundError(
+                    f"Item {model['id']} in collection {model['collection']}"
+                )
             # SQLAlchemy orm updates don't seem to like geoalchemy types
             db_model = self.item_serializer.stac_to_db(model)
             query.update(self.item_serializer.row_to_dict(db_model))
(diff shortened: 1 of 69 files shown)


stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
```diff
@@ -91,10 +96,15 @@ def delete_item(
         """Delete item."""
         base_url = str(kwargs["request"].base_url)
         with self.session.writer.context_session() as session:
+            query = session.query(self.item_table).filter(
```
lossyrob, line 99: This definition of `query` seems to be overridden by the next line. Perhaps use `query` instead of `session.query(...)` in the next line.

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
```diff
@@ -58,8 +58,13 @@ def update_item(self, model: stac_types.Item, **kwargs) -> stac_types.Item:
             query = session.query(self.item_table).filter(
                 self.item_table.id == model["id"]
             )
+            query = session.query(self.item_table).filter(
```
lossyrob, line 61: This definition of `query` seems to override the next line. Perhaps use `query` instead of `session.query(...)` here.

lossyrob (changes requested):
Added some comments.

Also this PR should have tests that prove the update and delete work in the case where two collections have the same Item ID, which I think would have caught the issues I pointed out.

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py
jonhealy1: True thanks

jonhealy1:
> Added some comments.
> 
> Also this PR should have tests that prove the update and delete work in the case where two collections have the same Item ID, which I think would have caught the issues I pointed out.

I will write some tests

jonhealy1:
@lossyrob Thanks for the feedback - I think the tests are good ...

moradology (approved):
I've gone through the changes here and they look good to me - @lossyrob were there any amendments you had in mind?

geospatial-jeff (approved):
LGTM, thanks @jonhealy1

stac_fastapi/sqlalchemy/tests/conftest.py
```diff
@@ -152,3 +152,13 @@ def app_client(api_client, load_test_data, postgres_transactions):
 
     with TestClient(api_client.app) as test_app:
         yield test_app
+
+
+@pytest.fixture
+def app_client_2(api_client, load_test_data, postgres_transactions):
```
geospatial-jeff, line 158: Actually, small comment.  Instead of creating a new app client fixture, I think it would be better if we create the second collection in each test case (`test_create_item_duplicate`, `test_delete_item_duplicate`, and `test_update_item_duplicate`).

jonhealy1: @Jeff good call. I have done it, thanks!

geospatial-jeff (approved):

lossyrob (approved):

PR #307, opened 2021-12-09
Bump pgstac version to 0.4.0

moradology:
**Related Issue(s):** #283

**Description:**
This PR upgrades the pgstac dependency to 0.4.0 which includes a number of bug fixes, performance improvements, and support for cql2.

**PR Checklist:**

- [x] Code is formatted and linted (run `pre-commit run --all-files`)
- [x] Tests pass (run `make test`)
- [x] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/master/CHANGES.md).

PR #305, opened 2021-12-02
Updated PgstacSearch to include two new optional properties for pgstac v0.4.0

edkeeble:
**Related Issue(s):** #

**Description:**

**PR Checklist:**

- [x] Code is formatted and linted (run `pre-commit run --all-files`)
- [x] Tests pass (run `make test`)
- [ ] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [ ] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/master/CHANGES.md).

Issue #283, opened 2021-11-03
Upgrade pgstac version to v0.3.5 or v0.3.6


zstatmanweil:
There are some key performance improvements and updates in the recent versions of pgstac (e.g., being able to turn off "context") that would be great to use with stac-fastapi. Not sure if there are breaking changes in these new versions of pgstac, or the version pinned can just be upgraded. Thanks!

moradology:
Looks like this one got lost in all the shuffle. 0.4.0 came out a few days ago so hopefully we can just jump to that

PR #308, opened 2021-12-13
Pgstac v0.4.0b

bitner:
**Related Issue(s):** #283
This PR should supersede #307 and #305 

**Description:**
 - Upgrades pgstac backend to 0.4.0
 - Enables cql2-json
   - pgstac enables as default (settable in pgstac configuration)
   - can be set per query using "filter-lang" parameter ("cql-json" or "cql2-json")
   - stac-fastapi-pgstac will automatically use cql-json if query, bbox, ids, collections, or datetime parameter are set for backwards compatibility

**PR Checklist:**

### DIFF
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py
@@ -1,6 +1,6 @@
 """stac_fastapi.types.search module."""
 
-from typing import Optional
+from typing import Dict, Optional
 
 from pydantic import validator
 
@@ -14,6 +14,7 @@ class PgstacSearch(BaseSearchPostRequest):
     """
 
     datetime: Optional[str] = None
+    conf: Optional[Dict] = None
 
     @validator("datetime")
     def validate_datetime(cls, v):
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py
@@ -20,3 +20,24 @@ class PgstacSearch(BaseSearchPostRequest):
     def validate_datetime(cls, v):
         """Pgstac does not require the base validator for datetime."""
         return v
+
+    @validator("filter_lang", pre=False, check_fields=False, always=True)
+    def validate_query_uses_cql(cls, v, values):
+        """If using query syntax, forces cql-json."""
+        retval = v
+        if values.get("query", None) is not None:
+            retval = "cql-json"
+        if values.get("collections", None) is not None:
+            retval = "cql-json"
+        if values.get("ids", None) is not None:
+            retval = "cql-json"
+        if values.get("datetime", None) is not None:
+            retval = "cql-json"
+        if values.get("bbox", None) is not None:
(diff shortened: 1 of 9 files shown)


bitner:
@moradology @edkeeble This PR includes the changes from #305 and #307 adding some basic tests to make sure that cql-json and cql2-json are both working. Would be great if you could throw in a review.

stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py
```diff
@@ -14,8 +14,30 @@ class PgstacSearch(BaseSearchPostRequest):
...
+        if values.get("datetime", None) is not None:
+            retval = "cql-json"
+        if values.get("bbox", None) is not None:
+            retval = "cql-json"
+        if v == "cql2-json" and retval == "cql-json":
+            raise ValueError(
+                "query, collections, ids, datetime, and bbox"
+                "parameters are not available in cql2-json"
+            )
+        return retval
```
edkeeble, line 43: `cql-text` and `cql-json` share the parameters listed here, but if `filter_lang` is "cql-text", it will be transformed to "cql-json". Is that the intention?

We might be able to condense this and avoid some duplicate code with the following snippet. It does change the logic slightly, though:

* It will throw if any CQL parameter is provided, even if its value is `null`. e.g. `{"filter-lang": "cql2-json", "query": null}` will throw the ValueError. I feel like that is still correct.
* It will only default to `cql-json` if `filter-lang` isn't specified (i.e. it will preserve `cql-text`)

```
cql_params = ("query", "collections", "ids", "datetime", "bbox")
if any(cql_param in values for cql_param in cql_params):
    if v == FilterLang.cql2_json:
        raise ValueError(f"{', '.join(invalid_cql2_params[:-1])} and {invalid_cql2_params[-1]} parameters are not available in cql2-json")
    else:
        return v or FilterLang.cql_json  # if `filter-lang` isn't specified and CQL parameters are provided, default to `cql-json`
return v
```

moradology (approved):
Nice. And thanks for the tests. LGTM 👍

edkeeble (approved):
Looks good to me. Thanks for wrapping this up @bitner!

I had one question around the `filter_lang` validator and whether we want to transform `cql-text` to `cql-json`, as well as a suggestion for avoiding some duplicate code, but it isn't critical. If the validator works as you intended, then it's good with me.

stac_fastapi/pgstac/stac_fastapi/pgstac/types/search.py
bitner: PGStac only accepts json, so in the case of a GET request, that request is just modified into CQL json.  The ultimate plan (which I will start working on this week, but probably won't finish before the holidays) for stac-fastapi-pgstac will be to add a cql2-json output to pygeofilter and to use pygeofilter to translate cql-text or cql2-text into cql2-json and then send that to pgstac.

vincentsarago:
Is there anything here that should be added in the changelog @moradology @bitner ?

moradology:
Probably not a bad idea to mention the bump and CQL2 support. I'll toss up a PR for that. Thanks for keeping an eye out

PR #318, opened 2022-01-11
Change of datetime type in BaseSearchGetRequest + test improvement

### LINKED ISSUE #315: Datetime no longer a parameter option in GET /search
In a recent [PR](https://github.com/stac-utils/stac-fastapi/pull/308/files),`Dict` was added as a type option for datetime in `BaseSearchGetRequest` 
https://github.com/stac-utils/stac-fastapi/blob/2f8e96bbdbad588e3b5df7f28dda5bd737228fcc/stac_fastapi/types/stac_fastapi/types/search.py#L76

After this change, for some reason `datetime` is no longer registering as a parameter in the GET /search endpoint for sqlalchemy or pgstac backend. You can confirm this by running it locally, and looking at the `/docs` or `/api` pages.  Alternatively, you can run a GET request on this URL and see that the item returned should actually be filtered out based on the datetime:

```http://0.0.0.0:8082/search?collections=joplin&ids=f2cca2a3-288b-4518-8a3e-a4492bb60b08&datetime=2021-02-02T00:00:00Z/2022-02-02T00:00:00Z```

I removed the `Dict` option and datetime reappeared as a parameter. 

Since this was made as part of the pgstac upgrade to 0.4.0, I assume it has something to do with that, but downstream all the functions expect a `str` or `datetime` so I haven't been able to figure it out. I see that cql2-json can take a dictionary for datetime within the filter parameter, but since the `datetime` parameter is not available in cql2-json, I don't think that applies here.

zstatmanweil:
**Related Issue(s):** #315 

**Description:**
This closes #315 . I made a minor change to fix that issue of removing `Dict` as a type option for `datetime` in `BaseSearchGetRequest` after a lot of investigation to determine it doesn't ever appear as a `Dict` downstream. Most of the PR includes expansion of the existing tests. 

The current tests weren't catching the issue described in #315 for a couple reasons: 1) some of the tests were testing too many things at once. For example, the temporal tests targeted at a `datetime` search, also included a `bbox` or `intersects` parameter; and 2) in many of the tests, only one item was being added to the test collection and then searched so when a parameter wasn't properly filtering the results (like in this case), it wasn't being caught. 

**PR Checklist:**


bitner:
Thanks for all the tests @zstatmanweil ! This looks good to me.

moradology (approved):

PR #309, opened 2021-12-14
Remove titiler demo extension

moradology:
**Related Issue(s):** #175 #245

**Description:**
This PR removes the `tiles` extension as well as the associated links which are currently added regardless of that extension being turned on. 

**PR Checklist:**

- [x] Code is formatted and linted (run `pre-commit run --all-files`)
- [x] Tests pass (run `make test`)
- [x] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [ ] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/master/CHANGES.md).

PR #330, opened 2022-01-14
Update docker compose for sqlalchemy

jonhealy1:
Unless I'm missing something, these variables are only needed for pgstac?

**Related Issue(s):** #

**Description:**
Update docker-compose for sqlalchemy to remove variables that are only needed for pgstac - I think. 

**PR Checklist:**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,9 +19,6 @@ services:
       - POSTGRES_HOST_WRITER=database
       - POSTGRES_PORT=5432
       - WEB_CONCURRENCY=10
-      - VSI_CACHE=TRUE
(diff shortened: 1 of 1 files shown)


geospatial-jeff:
We could also get rid of the GDAL environment variables, they are only needed for the tiles extension which was removed in https://github.com/stac-utils/stac-fastapi/pull/309

jonhealy1:
> We could also get rid of the GDAL environment variables, they are only needed for the tiles extension which was removed in #309

Does this mean that pgstac is not using gdal anymore either? The pgstac tests pass without them.

moradology (approved):
Tested and things seem to work!

Issue #312, opened 2022-01-04
Children extension


moradology:
In the upcoming (1.0.0-beta.5) release of stac-spec, a few new extension conformance classes will be added. Among them is an extension which will add a `/children` endpoint. Quoting the tentative stac-spec document,
"A STAC API Landing Page (a Catalog) can return information about the Catalog and Collection child objects it contains using the link relation children to an endpoint /children. The purpose of this endpoint is to present a single resource from which clients can retrieve all the immediate child objects of a Catalog, which may be Catalog or Collection objects. The child link relations in a Catalog already allow for describing these relationships, but require a client to retrieve each resource URL to find any information about the children (e.g., title, description), which can cause significant performance issues in user-facing applications."
More, here: https://github.com/radiantearth/stac-api-spec/blob/5a694d5daf838e5edb974a2e7fdfa59547435340/children/README.md

jonhealy1:
@moradology Does this mean that there will now need to be a /catalogs route? Every catalog now is a separate api instance I think - is this changing? This may not be as complicated as it seems ...

moradology:
This new conformance class should be read as *only* an extension to core and is not required to host a valid STAC API which advertises the `core` but not the `children` conformance class. Each catalog can still be treated as a separate API instance but, assuming this extension is implemented, those catalogs with other catalogs immediately under them can be expected to advertise said sub-catalogs and any subordinate collections at a new `<super-catalog>/children` endpoint

jonhealy1:
@moradology Ok I think I understand what you are saying. Right now a root catalog is created automatically and child links are added by calling the /collections route. Someone would need to be able to inject catalog links in to the root catalog? A root catalog would have to be told what catalogs exist underneath it or would those catalogs only be displayed in the /children endpoint? I think it's a good idea.

Issue #311, opened 2022-01-04
Browsable extension


moradology:
In the upcoming (1.0.0-beta.5) release of stac-spec, a few new extension conformance classes will be added. Among them is an extension which guarantees that all Items can be accessed by clicking through `item` and `children` links. More, here: https://github.com/radiantearth/stac-api-spec/blob/5a694d5daf838e5edb974a2e7fdfa59547435340/browseable/README.md

PR #336, opened 2022-01-25
Browseable and Children features from 1.0.0-beta5

moradology:
**Related Issue(s):** #311 #312

**Description:**
This PR adds a workaround which is intended to allow `children` and `browsable` conformance classes to be enabled in stac-fastapi while deeper changes for a more robust implementation are being developed in pgstac. For a sense of what specifying browsable hierarchies looks like, pay special attention to `hierarchy.json` and `hierarchy.py`

Additionally, clients have been broken out of the hundreds-of-lines file they used to live in and organized into their own package.

image
image
image

**PR Checklist:**


moradology:
With the most recent commit (191b75f), each `catalog` endpoint is, itself, a STAC API. Reviewers ought to keep an eye out for code that can be deduplicated without losing important context (e.g. a try/catch block which issues a 404 on exception is quite likely more useful duplicated as it doesn't hide relevant information for understanding an endpoint's behavior)

Here's an ancient meme that seems especially relevant:
image

moradology:
With the most recent commit (191b75f), each catalog endpoint is, itself, a STAC API. Reviewers ought to keep an eye out for code that can be deduplicated without losing important context (e.g. a try/catch block which issues a 404 on exception is quite likely more useful duplicated as it doesn't hide relevant information for understanding an endpoint's behavior)

stac_fastapi/types/stac_fastapi/types/clients/async_core.py
```diff
@@ -0,0 +1,592 @@
...
+
+        Returns:
+            ItemCollection containing items which match the search criteria.
+        """
+        # Pydantic is fine for specifying body parameters but proves difficult to
+        # use for Path parameters. Here, the starlette request is inspected instead
+        request_path = kwargs["request"]["path"]
+        split_path = request_path.split("/")[2:-1]
+        remaining_hierarchy = self.hierarchy_definition.copy()
+        # PGSTAC TODO: Add optional 'catalog' parameter to search, which will determine which
```
geospatial-jeff, line 374: are any of these PGSTAC TODO laying around the PR important to resolve before we merge?

stac_fastapi/types/stac_fastapi/types/clients/async_core.py
```diff
@@ -0,0 +1,592 @@
...
+            raise HTTPException(
+                status_code=404, detail=f"Catalog at {'/'.join(split_path)} not found"
+            )
+        child_collections = [
+            node["collection_id"]
+            for node in selected_catalog["children"]
+            if "collection_id" in node
+        ]
+        search_request.collections = child_collections
+        return await self.post_search(search_request, **kwargs)
```
geospatial-jeff, line 387: I'm glad this is just a wrapper on top of the existing `post_search`, super cool!

stac_fastapi/types/stac_fastapi/types/hierarchy.py
```diff
@@ -0,0 +1,200 @@
...
+        stac_version=stac_version,
+        conformsTo=conformance_classes,
+        links=standard_links + children_links + item_links,
+        stac_extensions=extension_schemas,
+    )
+    return catalog_page
+
+
+# PGSTAC TODO: function to allow ingest of hierarchy
+def parse_hierarchy(d: dict) -> BrowseableNode:
```
geospatial-jeff, line 181: Return type should be a Union of all three nodes, but this function is only ever called on a dictionary that would return a `BrowseableNode`.  So I'm not sure if the return type is wrong or the `CollectionNode` and `CatalogNode` returns aren't needed.

geospatial-jeff (changes requested):

stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
```diff
@@ -103,6 +109,53 @@ async def get_collection(self, collection_id: str, **kwargs) -> Collection:
 
         return Collection(**collection)
 
+    async def get_root_children(self, **kwargs) -> Children:
```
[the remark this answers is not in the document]
moradology: Agreed. In this PR it already got to the point that organization in a single file was getting absurd.

stac_fastapi/testdata/joplin/hierarchy.json
```diff
@@ -0,0 +1,35 @@
+{
```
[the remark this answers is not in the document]
moradology: Your scalability concern is reasonable. I'm having a hard time imagining it being a performance bottleneck, but JSON is brittle and it is (as currently imagined) directly maintained by developers (often a recipe for trouble). I'm not 100% sure what the best approach is to avoid such a difficulty. Perhaps rules for generating trees rather than fully specified trees? I'm open to suggestions.

Runtime addition of nodes should be possible with a small set of modifications. I agree that's important to add and am also in support of adding it in future updates

stac_fastapi/types/stac_fastapi/types/clients/async_core.py
moradology: These were initially added to provide some pointers to @bitner about what kinds of functionality could move back into pgstac, but I'm happy to get rid of them!

stac_fastapi/types/stac_fastapi/types/hierarchy.py
moradology: It is very possible that I'm missing something peculiar about python types or `TypedDict` inheritance, but I think we can avoid the `Union` here because of the inheritance structure - both `CollectionNode` and `CatalogNode` are subsumed under the `BrowseableNode` type. Let me know if I'm off base in this assertion

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -83,9 +86,15 @@ class StacApi:
     api_version: str = attr.ib(default="0.1")
     stac_version: str = attr.ib(default=STAC_VERSION)
     description: str = attr.ib(default="stac-fastapi")
+    search_get_request_base_model: Type[BaseSearchGetRequest] = attr.ib(
```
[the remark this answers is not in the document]
moradology: Yeah, this is a potentially ugly and maybe a point where you could provide some counsel. Basically, the need to supply a catalog URI parameter changes the constructed search model  so what I'm doing here is passing in the base `search_get_request_base_model` because it might possibly differ from the fully constructed `search_get_request_model` used by standard searches. Perhaps the answer is to just construct both of these inside of `core.py`. Let me know what you think.

gadomski:
@moradology is this workaround still necessary? I'm doing a little housekeeping and don't want to try and resolve these merge conflicts if its not needed.

moradology:
Hey @gadomski, I don't know that the children/browseable features are 100% necessary so it is probably not absolutely necessary that this workaround be included in the project. I'll close the PR for now but keep the branch around in case people want to refer back for a future implementation

Issue #340, opened 2022-01-27
Support runtime CORS configuration


captaincoordinates:
The [documentation](https://github.com/stac-utils/stac-fastapi/blob/5dc8e0eb63d9cc07663968ed5ed9505e753bf298/docs/tips-and-tricks.md#get-stac-fastapi-working-with-cors) describes how to configure CORS for stac-fastapi and this currently requires code modifications. As a result anyone wanting to run the API with CORS needs to fork the repo to add a few lines of code. This is not ideal as CORS is a common requirement.

stac-fastapi should support runtime configuration of CORS with the following approach:
- CORS configuration file added to a container image or copied / mounted into a running container
- environment variable points to file location
- API recognizes and loads CORS configuration during init sequence

This approach has been implemented in a fork and a PR will be submitted shortly.

vincentsarago:
> As a result anyone wanting to run the API with CORS needs to fork the repo

stac-fastapi is a proper python module, users should not have to fork but just do `pip install`. Then it's pretty easy to setup your own application: https://github.com/developmentseed/eoAPI/blob/master/src/eoapi/stac/eoapi/stac/app.py

I worry that if we add a special case for CORS, then another user might come with other middleware needs (e.g cache control). 

previous conversation: https://github.com/stac-utils/stac-fastapi/issues/127

🤦 well in ☝️ @geospatial-jeff makes a good point that cors should be maybe set by default and to be permissive (https://github.com/radiantearth/stac-api-spec/blob/master/implementation.md#cors) 

Still IMO, we shouldn't add more complexity on the `application` part, but work on the documentation explaining how create/customize stac-fastapi application

captaincoordinates:
@vincentsarago I appreciate your input, however I would counter that creating your own application is similarly overkill when all you want is a CORS configuration. Whether it's a forked repo or a new application, either way you're maintaining a repo purely for CORS. pgstac and sqlalchemy each contains >3,000 lines of python, which suggests that this is not a trivial exercise.

I'm not clear on why it's bad to set a precedent here - if it lowers the barriers to using the software without intrusive changes in the code I don't see a problem.

geospatial-jeff:
> pgstac and sqlalchemy each contains >3,000 lines of python, which suggests that this is not a trivial exercise.

[...]

Issue #127, opened 2021-04-27
stac-fastapi does not provide CORS header (Cross-Origin Resource Sharing)


ghost:
Hi all,

at first thanks for this great software!

When using "STAC browser" on a catalog created with stac-fastapi I get an error message that normally indicates that cors headers are not present.

"NetworkError when attempting to fetch resource.
Please note that some servers don't allow external access via web browsers (e.g., when CORS headers are not present).
Errored URL: https://localhost:8081"

Any comment is appreciated! If you point me to a location in the source code I can also try to include it.

vincentsarago:
@DanielMaierLGLN I think this is expected, It's let to the end users to `customize` the application layer. 

Basically you need to update https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/app.py#L31 as 👇 

```python
"""FastAPI application."""
from stac_fastapi.api.app import StacApi
from stac_fastapi.extensions.core import (
    FieldsExtension,
    QueryExtension,
    SortExtension,
    TransactionExtension,
)
from stac_fastapi.extensions.third_party import BulkTransactionExtension
from stac_fastapi.sqlalchemy.config import SqlalchemySettings
from stac_fastapi.sqlalchemy.core import CoreCrudClient
from stac_fastapi.sqlalchemy.session import Session
from stac_fastapi.sqlalchemy.transactions import (
    BulkTransactionsClient,
    TransactionsClient,
)

from starlette.middleware.cors import CORSMiddleware

settings = SqlalchemySettings()
session = Session.create_from_settings(settings)
api = StacApi(
    settings=settings,
    extensions=[
        TransactionExtension(client=TransactionsClient(session=session)),
        BulkTransactionExtension(client=BulkTransactionsClient(session=session)),
        FieldsExtension(),
        QueryExtension(),
        SortExtension(),
    ],
    client=CoreCrudClient(session=session),
)
app = api.app

# ADD CORSMiddleware
app.add_middleware(
    CORSMiddleware,
    allow_origins=["*"],
    allow_credentials=True,
    allow_methods=["GET", "POST],
    allow_headers=["*"],
)
```

DanielMaierLGLN:
Thank you very much @vincentsarago!
If it is expected to be done by the end users - it's ok for me (maybe it makes sense it include it in the documentation?).

I confirm that your solution is working (apart from a small typo in the app.add_middleware part).
[...]

PR #341, opened 2022-01-27
Enable runtime CORS configuration via environment variables

captaincoordinates:
**Related Issue(s):**
#340 

**Description:**
Support runtime CORS configuration via JSON configuration and environment variable. Missing environment variable, missing config file, un-parseable config file all result in no CORS configuration. Runtime CORS configuration will be ignored if the application has already added `CORSMiddleware`.

New tests have been added to ensure this functionality works as designed. Minor changes to testing infrastructure to support setup behaviour prior to some pytest fixtures running.

Documentation updated to recommend this approach to CORS configuration over the previous code modification approach.

**PR Checklist:**


geospatial-jeff:
@captaincoordinates Thanks for the PR! 🙏 

This feels a little overloaded.  Using a JSON file to store CORS configuration is a code leak and overly complex, typically this sort of configuration is the responsibility of the deployment and not the application code.  I would much rather reuse the FastAPI CORS middleware as-is, parameterizing the middleware settings w/ environment variables.  This provides a more flexible and robust way to configure/deploy the app and aligns better with microservice best practices (see https://12factor.net/config). 

Baking CORS into stac-fastapi does resolve #340, as a user will be able to configure CORS in stac-fastapi without forking the code or implementing their own app.  Since the stac-api spec does recommend enabling CORS, I think it is totally fine to include it as a default middleware (by appending to `StacApi.middlewares` (https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/stac_fastapi/api/app.py#L90).  No need to implement our own custom `MiddlewareConfig` since FastAPI already provides a way to add middleware to the application.

This will close #127 as well.

vincentsarago:
we could simply do something like 

```python
from starlette.middleware import cors

class CORSMiddleware(cors.CORSMiddleware):
    """Starlette CORS Middleware with default."""

    def __init__(
        self,
        app: ASGIApp,
        allow_origins: Sequence[str] = ("*",),
        allow_methods: Sequence[str] = ("OPTIONS, POST, GET",),
        allow_headers: Sequence[str] = ("Content-Type",),
        allow_credentials: bool = False,
        allow_origin_regex: str = None,
        expose_headers: Sequence[str] = (),
        max_age: int = 600,
    ) -> None:
        """Create CORSMiddleware Object."""
        super().__init__(
            app,
            allow_origins=allow_origins,
            allow_methods=allow_methods,
            allow_headers=allow_headers,
            allow_credentials=allow_credentials,
            allow_origin_regex=allow_origin_regex,
            expose_headers=expose_headers,
            max_age=max_age,
        )

    # https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/stac_fastapi/api/app.py#L90
    middlewares: List = attr.ib(
        default=attr.Factory(lambda: [BrotliMiddleware, CORSMiddleware]),
    )
```

Note: each backend application will need to update the `CORSMiddleware` if they need transaction methods (https://github.com/stac-utils/stac-fastapi/blob/5dc8e0eb63d9cc07663968ed5ed9505e753bf298/stac_fastapi/types/stac_fastapi/types/core.py#L25)

tomfumb:
@geospatial-jeff @vincentsarago thanks for the feedback. If you're each happy for CORS to be a default middleware I will follow @vincentsarago's approach above with the addition of env var configuration for each of the options, defaulting to the values provided in that example. BRB with an updated PR

captaincoordinates:
@geospatial-jeff @vincentsarago thanks for the feedback. If you're each happy for CORS to be a default middleware I will follow @vincentsarago's approach above with the addition of env var configuration for each of the options, defaulting to the values provided in that example. BRB with an updated PR

captaincoordinates:
PR updated with a variation on the suggested approach to permit configuration via environment variables. Some additional tests are required for the various CORS configuration options but I was hoping to get initial feedback on the overall approach before committing that effort (hence draft PR status).

captaincoordinates:
I'm working on aligning env var configuration with the ApiSettings sub-class approach after realising that there was a good convention already in place

stac_fastapi/api/stac_fastapi/api/config.py
```diff
@@ -22,3 +28,48 @@ class AddOns(enum.Enum):
...
+
+def env_to_int(env_var: str, default: int) -> int:
+    """Retrieve an int from an env var, or default if missing."""
+    if env_var in environ:
+        value = environ[env_var].strip()
+        if value.isdigit():
+            return int(value)
+        else:
+            logger.warning(f"{env_var} set but not a valid int")
+    return default
```
vincentsarago, line 75: could we just use `pydantic.BaseSettings` https://pydantic-docs.helpmanual.io/usage/settings/

It should take care of everything needed

captaincoordinates: Yes, sorry, I just had that realisation a few minutes earlier

stac_fastapi/api/stac_fastapi/api/__init__.py
```diff
@@ -1 +1,6 @@
 """api submodule."""
+from typing import Final
+
+from stac_fastapi.api.config import Settings
+
+settings: Final = Settings()
```
vincentsarago, line 6: why do we need this ?

stac_fastapi/pgstac/tests/api/test_api.py
```diff
@@ -302,3 +316,54 @@ async def test_search_line_string_intersects(
...
+    assert resp.status_code == HTTPStatus.OK
+    assert "access-control-allow-origin" not in resp.headers
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("app_client", [{"setup_func": cors_deny}], indirect=True)
+async def test_with_mismatch_cors_origin(app_client):
+    resp = await app_client.get("/", headers={"Origin": cors_origin_1})
+    assert resp.status_code == HTTPStatus.OK
+    assert "access-control-allow-origin" not in resp.headers
```
vincentsarago, line 369: Do we need to test the middleware on the stac-fastapi backend? I guess we could just write simple tests using a minimalist fastapi application.

stac_fastapi/api/stac_fastapi/api/__init__.py
captaincoordinates: The goal is to have a single reference to `settings` that can be both accessed by the application and modified by tests for setup / teardown. That single reference could be implemented elsewhere, e.g. in config.py, but this approach is a personal preference for me as it makes clear that `settings`'s purview is the entire `stac_fastapi.api` package. Happy to change if desired.

stac_fastapi/pgstac/tests/api/test_api.py
captaincoordinates: I see this as following precedent - I don't see any tests for the `stac_fastapi.api` package, which suggests all "core" functionality is tested exclusively via the backends.

captaincoordinates:
FYI I'm working on the docs build failure. I checked `make test` before pushing but not `make docs`

stac_fastapi/api/stac_fastapi/api/__init__.py
captaincoordinates: Looks like I have to move it anyway for `make docs-image` to succeed. It's failing because pydantic is not available when the build first tries to import the stac_fastapi.api package.

stac_fastapi/pgstac/tests/api/test_api.py
vincentsarago: can we do like https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/tests/test_api.py instead of duplicating those tests

captaincoordinates: @vincentsarago I moved the tests you referenced to a more meaningful filename and extracted to shared code the necessary pieces for my tests to run on api.

moradology:
@captaincoordinates Curious about the status on this

captaincoordinates:
@moradology PR feedback went quiet, so I'm not sure

moradology:
Cool, I've just been away from the lib for a bit and am trying to get back on top of the PRs I've missed! I'll poke this today to keep it moving!

moradology:
@captaincoordinates Hey Tom, so I've got a rebased branch [here](https://github.com/moradology/stac-fastapi/tree/feature/1) which you might look at and force onto your @sparkgeo `feature/1` branch (I appear to not have the permissions to do this). Let me know if I can be of further assistance

captaincoordinates:
@moradology thanks, I'm pinned by project work at the moment but I will hopefully get to it during the next few days

captaincoordinates:
@moradology should be good to go. Thanks for fixing those test errors, for a moment there I thought I was going to have to figure it out myself

stac_fastapi/api/stac_fastapi/api/config.py
```diff
@@ -22,3 +28,18 @@ class AddOns(enum.Enum):
     """Enumeration of available third party add ons."""
 
     bulk_transaction = "bulk-transaction"
+
+
+class Settings(BaseSettings):
```
lossyrob, line 33: The fastapi app that comes packaged in stac-fastapi isn't necessary to use the project, and in the past has been talked about as a convenience rather than a core feature of the library (given that a client project perhaps should create their own App to customize with any middleware particular to the application, etc).

Since these settings apply to the fastapi app, perhaps we should give them a name that's less generic - e.g. a reader might thing that these are the most core settings, vs "ApiSettings".

Perhaps `FastApiAppSettings`?

Same with the exported instance below.

lossyrob (changes requested):

stac_fastapi/api/stac_fastapi/api/config.py
captaincoordinates: @lossyrob thanks for the good suggestion. I've been unable to look at this for a while but thanks to @moradology this has been implemented

jlev:
Any chance this will land soon? Would love to get CORS headers configured the right way with environment variables, instead of hacking on a fork of the code.

captaincoordinates:
@jlev I don't know why this has been hanging for so long. Now there are merge conflicts that are difficult to resolve and I don't know when I'll be able to fix them

alexgleith:
Would love to see this merged. Happy to help out if I can.

geospatial-jeff:
I managed to resolve most of the merge conflicts but I don't think we should merge this PR as the same may be accomplished with a lot less code by forcing user to configure/attach middleware themselves.

```python
from fastapi.middleware.cors import CORSMiddleware

# Create you api.
api = StacApi(...)

# Add the middleware
# https://fastapi.tiangolo.com/tutorial/cors/
api.app.add_middleware(
    CORSMiddleware,
    allow_methods=["*"],
    allow_headers=os.getenv("CORS_ALLOW_HEADERS")
)
```

I see why the PR is written this way; adding of middlewares is currently done in the post-init and doesn't currently expose passing of positional args which requires overloading the middleware class itself to expose these arguments.  This is a bad pattern, and I'd rather fix the underlying pattern than overloading each middleware.  I think we should probably remove the `StacApi.middlewares` attribute all-together and write a `StacApi.add_middleware` method which just wraps `FastApi.add_middleware` and allows users to add their own middlewares w/ less abstraction.

geospatial-jeff:
Closing this in favor of #435 and #436

Issue #356, opened 2022-02-24
SQLAlchemy backend bulk item insert fails 


philvarner:
https://github.com/stac-utils/stac-fastapi/blob/79c0cee7795b95a9ffc5f350c62d4bf69fcc2875/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/transactions.py#L154

I'm not sure what the error is going to be here, because I found this with some derived code that failed in an arbitrarily different way. 

The problem is that `items` is an instance of `Items`, and using this in a list comprehension will iterate over the (string) keys in the `items` dict<string,Any> attribute of Items per it's implementation of `iter`, rather than each Item itself (the values of the dict) as this code expects.

I fixed this in my code with:

```
 processed_items = [self._preprocess_item(item, base_url) for item in items.items.values()]
```

geospatial-jeff:
Closed with #358.  BTW @philvarner I gave you write access as requested in that PR.

Issue #355, opened 2022-02-24
Bulk Transactions Items iterator returns keys instead of Items


philvarner:
https://github.com/stac-utils/stac-fastapi/blob/33611f7da569488641dac7bb23efd43feb219f98/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py#L21

in Bulk Transactions, the Items iterator returns an iterator over the keys in the `items` dictionary. This isn't very useful and (I would say) unexpected, as I'd expect to get an iterator of the Items objects instead (the values of that dict).

Issue #196, opened 2021-07-26
Update Input model for bulk insert or clarify docstring! 


vincentsarago:
The current input model for the bulk extension is a bit unclear to me. Reading 

https://github.com/stac-utils/stac-fastapi/blob/bd6f6613e2c3969d5a68fa33509c5b4deb3e7aa3/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py#L14-L21

It seems that we have to provide input in form of 
```
{
    "items": {
        "id1": { item1 },
        "id2": { item2 },
        "id3": { item1 }
    }
}
```

This is kinda weird, and I was expecting to just be able to profile something closer to `[item1, item2, item3]` 🤷 

If the `dict` form is going to stay, I think we should at least update the doc string (and remove the word `list`)

cc @geospatial-jeff

philvarner:
We could also remove this entirely, and add support (currently missing, and therefore out of compliance with the Transaction Extension) for POST /collection/{cid}/items with an ItemCollection body.

geospatial-jeff:
> We could also remove this entirely, and add support (currently missing, and therefore out of compliance with the Transaction Extension) for POST /collection/{cid}/items with an ItemCollection body.

I think this is a really good idea.

PR #358, opened 2022-02-26
fix sqlalchemy bulk item insertion

philvarner:
**Related Issue(s):** 

- #355
- #356
- #196

**Description:**

SQLAlchemy bulk item insertion failed because the `Items` iterator returns the keys for the dict it contains, rather than the Item objects themselves as it's docstring says it should (and that it would make sense to).

This fixes that problem, and documents what the format of the JSON object that is the input to the bulk insert endpoint is.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
@@ -12,13 +12,13 @@ from stac_fastapi.types.extension import ApiExtension
 
 
 class Items(BaseModel):
-    """A list of STAC items."""
+    """A group of STAC Item objects, in the form of a dictionary from Item.id -> Item."""
 
     items: Dict[str, Any]
 
     def __iter__(self):
-        """Return an iterable of STAC items."""
-        return iter(self.items)
+        """Return an iterable of STAC Item objects."""
+        return iter(self.items.values())
 
 
 @attr.s  # type: ignore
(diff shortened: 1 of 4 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
```diff
@@ -12,13 +12,13 @@
 
 
 class Items(BaseModel):
-    """A list of STAC items."""
+    """A group of STAC Item objects, in the form of a dictionary from Item.id -> Item."""
 
     items: Dict[str, Any]
```
kylebarron, line 17: I haven't followed all the discussion but do we need it as a dict? Or could we just have a list of stac items?

geospatial-jeff: a list of items would align better with `ItemCollection`

geospatial-jeff (approved):
Thanks for fixing the extension, let's open another issue if we want to switch this to `ItemCollection` as discussed in the linked issues.

geospatial-jeff:
Created an issue here: https://github.com/stac-utils/stac-fastapi/issues/362

stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
philvarner: Correct, I just wanted to fix the behavior rather than change the expected input

philvarner:
@geospatial-jeff looks like you'll need to merge, I don't have write perms.

Issue #365, opened 2022-03-01
Ambigous datetime values are accepted but should be rejected


philvarner:
These two datetime values are accepted for query, but are not valid RFC 3339 datetimes, and should be rejected with a 400.

- Search with datetime=1937-01-01T12:00:27.87+0100 returned status code 200 instead of 400 -- no colon between hour and minute in tz
- Search with datetime=1985-12-12T23:20:50.52 returned status code 200 instead of 400 -- no timezone

geospatial-jeff:
Closed with #368

Issue #204, opened 2021-08-05
Breaking Changes on STAC Item date


jaysnm:

geospatial-jeff:
I believe using pydantic's datetime parser should fix this instead of a single hard-coded datetime pattern.  We made the same change to datetime parsing in stac-pydantic for similar reasons (https://github.com/stac-utils/stac-pydantic/pull/75).

jaysnm:
Thanks Jeff

Sounds promising that you already have an idea for a sustainable fix.

jaysnm:
Hi @geospatial-jeff 

What do you think about `python-dateutil`? 
`dateutil.parser.parse` function takes care of date format conversion!

geospatial-jeff:
I think `python-dateutil` is fine, used it before in other projects.  But I'd prefer using pydantic's datetime parser in this situation to be consistent with the validation logic in stac-pydantic.  Overall I don't think either `python-dateutil` or `pydantic` are what we will settle on long term, there is more context in #161

philvarner:
I just added this to the STAC API impl guide also https://github.com/radiantearth/stac-api-spec/blob/master/implementation.md#datetime-parameter-handling

philvarner:
Also, this related issue https://github.com/stac-utils/stac-fastapi/issues/161

kylebarron:
I just came across [`ciso8601`](https://github.com/closeio/ciso8601). It looks to be both faster and more popular than `pyiso8601`

philvarner:
I just ran my test datetime values against it ciso8601, and only failed when it should have passed https://github.com/closeio/ciso8601/issues/110

philvarner:
I ran it against a few others:

```
ciso8601
Error: 1985-04-12T23:20:50,52Z did not parse. UTC offset is mandatory in RFC 3339 format.

arrow
Error: 1985-04-12t23:20:50.000z did not parse. Could not match input '1985-04-12t23:20:50.000z' to any of the following formats: YYYY-MM-DD, YYYY-M-DD, YYYY-M-D, YYYY/MM/DD, YYYY/M/DD, YYYY/M/D, YYYY.MM.DD, YYYY.M.DD, YYYY.M.D, YYYYMMDD, YYYY-DDDD, YYYYDDDD, YYYY-MM, YYYY/MM, YYYY.MM, YYYY, W.
Error: 1985-04-12 parsed, but should not have. result: 1985-04-12T00:00:00+00:00
[...]

Issue #161, opened 2021-06-18
Issues with Search datetime parameter parsing - blank start or end range and lowercase t or z


philvarner:
These issues were found on https://planetarycomputer.microsoft.com/api/stac/v1/ and locally running both the pgstac and sqlalchemy backends.

These `datetime` parameter values applied to /search should be accepted, but incorrectly return a 400:

```
Search with datetime=/1985-04-12T23:20:50.52Z returned status code 400
Search with datetime=1985-04-12T23:20:50.52Z/ returned status code 400
```

This is likely because only `..` is accepted as an open ended interval, whereas blank string is also allowed per: "Open ranges in time intervals at the start or end are supported using a double-dot (..) or an empty string for the start/end." http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameter_datetime

This datetime string gave a 400, likely because it uses `t` and `z`, which is acceptable.

```
Search with datetime=1985-04-12t23:20:50.000z returned status code 400
```

A large number of valid RFC3339 strings can be found here: https://github.com/stac-utils/stac-api-validator/blob/afb5a91ed6ce804d6b33fb304d6532dd4c50a227/stac_api_validator/validations.py#L38

geospatial-jeff:
The new version of stac-pydantic will use pydantic's datetime parser, which still doesn't quite do the job:

```python
from pydantic.datetime_parse import parse_datetime


dates = [
    "1985-04-12T23:20:50.52Z",
    "1985-04-12",
    "1937-01-01T12:00:27.87+0100",
    "37-01-01T12:00:27.87",
    "1985-12-12T23:20:50.52",
    "1985-04-12T23:20:50,Z"
]

for d in dates:
    try:
        parse_datetime(d)
    except:
        print("Invalid: ", d)
        continue
    print("Valid: ", d)

"Valid:  1985-04-12T23:20:50.52Z"
"Invalid:  1985-04-12"
"Valid:  1937-01-01T12:00:27.87+0100"
"Invalid:  37-01-01T12:00:27.87"
"Valid:  1985-12-12T23:20:50.52"
"Invalid:  1985-04-12T23:20:50,Z"
```

Looks like most of the cases that break are related to the timezone requirements:
> - a timezone offset must have a colon between the HH and MM
> - datetimes must have a timezone (Z or [+-]HH:MM)

[...]

PR #368, opened 2022-03-02
Improve RFC 3339 datetime handling

philvarner:
**Related Issue(s):** 

- #161 
- #365 
- #204
- https://github.com/stac-utils/stac-pydantic/issues/78

**Description:**

**PR Checklist:**

### DIFF
--- a/Dockerfile.docs
+++ b/Dockerfile.docs
@@ -1,5 +1,6 @@
 FROM python:3.8-slim
 
+RUN apt update && apt install -y build-essential
 RUN python -m pip install --upgrade pip
 RUN python -m pip install mkdocs mkdocs-material pdocs
 
--- a/Dockerfile.docs
+++ b/Dockerfile.docs
@@ -1,6 +1,8 @@
 FROM python:3.8-slim
 
+# build-essential is required to build a wheel for ciso8601
 RUN apt update && apt install -y build-essential
+
 RUN python -m pip install --upgrade pip
 RUN python -m pip install mkdocs mkdocs-material pdocs
 
--- a/Dockerfile.docs
+++ b/Dockerfile.docs
@@ -10,12 +10,11 @@ COPY . /opt/src
 
 WORKDIR /opt/src
 
-RUN python -m pip install -e \
+RUN python -m pip install \
     stac_fastapi/api \
     stac_fastapi/types \
     stac_fastapi/extensions \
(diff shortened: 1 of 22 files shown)


stac_fastapi/types/stac_fastapi/types/search.py
```diff
@@ -171,18 +159,22 @@ def validate_datetime(cls, v):
 
         dates = []
         for value in values:
-            if value == "..":
-                dates.append(value)
+            if value == ".." or value == "":
```
philvarner, line 162: this should probably be replace with parse_interval also

jonhealy1 (approved):

stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/serializers.py
```diff
@@ -1,17 +1,17 @@
 """Serializers."""
 import abc
 import json
-from datetime import datetime
 from typing import TypedDict
 
 import attr
 import geoalchemy2 as ga
-from stac_pydantic.shared import DATETIME_RFC339
```
philvarner, line 9: DATETIME_RFC339 has been deprecated in stac_pydantic. It has several problems including, not parsing fractional seconds or timezones other than Z.

Dockerfile.docs
```diff
@@ -1,19 +1,21 @@
...
+
 RUN python -m pip install --upgrade pip
 RUN python -m pip install mkdocs mkdocs-material pdocs
 
 COPY . /opt/src
 
 WORKDIR /opt/src
 
-RUN python -m pip install -e \
+RUN python -m pip install \
```
philvarner, line 13: `-e` isn't needed in this case, and it only affects the first package name after it (rather than all of them)

stac_fastapi/types/setup.py
```diff
@@ -10,6 +10,8 @@
     "attrs",
     "pydantic[dotenv]",
     "stac_pydantic==2.0.*",
+    "pystac==1.*",
```
philvarner, line 13: it took me 2 hours to figure out that the docs broke because I'd added an import from pystac, but it's not a dependency of this package. The tests worked because everything gets installed when they run.

moradology:
Looks like a black update broke a whole bunch of builds out there: https://github.com/psf/black/issues/2964

moradology (approved):

Issue #371, opened 2022-03-05
Conformance classes inconsistency


m-mohr:
Some conformance classes have a trailing slash (sometimes before a fragment) and sometimes not.
For example sort is `https://api.stacspec.org/v1.0.0-beta.4/item-search/#sort` although it seems from the spec that it should be `https://api.stacspec.org/v1.0.0-beta.4/item-search#sort`

m-mohr:
There seem to be a couple more things inconsistent: https://github.com/stac-utils/stac-fastapi/pull/417

geospatial-jeff:
I think this may have been fixed but I'd like to do one more pass through all of the conformance classes after #432 is finished.

gadomski:
I checked both backends locally and they look fine. I'm going to close this a fixed pre-2.4.4, please reopen if there's incorrect conformance classes found.

m-mohr:
@gadomski Looks better now, but I just found that at least MSPC still uses strange CQL Text/JSON conformance classes.
Not sure whether that's a MS PC but or a stac fastapi bug though.

- MSPC: `https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:cql-text` and json equivalent
- The Filter extension indicates: `http://www.opengis.net/spec/cql2/1.0/conf/cql2-text` and json equivalent

gadomski:
Yup, I _think_ you're right but will ask Phil to check to make sure (over on the new issue).

Issue #366, opened 2022-03-01
Update stac-api-spec version to 1.0.0-rc.1 and add collections conformance class.


philvarner:
1. spec can be update from 1.0.0-beta.4 to 1.0.0-rc.1. There were no substantive changes in beta.5 and rc.1 that require code changes (apart from the handful of issues that were previously out of compliance with the spec ([labeled with spec](https://github.com/stac-utils/stac-fastapi/labels/spec))
2. https://api.stacspec.org/v1.0.0-rc.1/collections can be added as a conformance class, as this is a subset of the already-implmented `STAC API - Features` conformance class.

geospatial-jeff:
Closed with #383

Issue #375, opened 2022-03-16
Transaction Extension missing support for PATCH /collections/{collectionId}/items/{featureId} endpoint


philvarner:
See:
https://github.com/radiantearth/stac-api-spec/tree/master/ogcapi-features/extensions/transaction

gadomski:
I'm moving this out of 2.4.4 as its non-essential and might not be widely used.

nleconte-csgroup:
Hello, we are interested in this feature for both transaction and collection transaction extensions.

PR #383, opened 2022-03-25
update all conformance classes to 1.0.0-rc.1 and fix several extension cc URIs with extra forward-slash

philvarner:
**Related Issue(s):** 

- https://github.com/stac-utils/stac-fastapi/issues/371
- https://github.com/stac-utils/stac-fastapi/issues/366

**Description:**

- update all the conformance classes to be rc.1
- fix the few extension conformance classes that had an extra `/` before the `#`
- add the collections conformance class 

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,24 +2,27 @@
 
 ## Unreleased
 
-* Update error response payloads to match the API spec. ([#361](https://github.com/stac-utils/stac-fastapi/pull/361))
-
 ### Added
 
 * Add hook to allow adding dependencies to routes. ([#295](https://github.com/stac-utils/stac-fastapi/pull/295))
+* Add STAC API - Collections conformance class. ([383](https://github.com/stac-utils/stac-fastapi/pull/383))
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


geospatial-jeff (approved):
LGTM!  We haven't implemented the changes to transactions extension (#384 and #385) but I don't think that is a big deal (https://github.com/radiantearth/stac-api-spec/blob/master/CHANGELOG.md#changed).

philvarner:
> LGTM! We haven't implemented the changes to transactions extension (#384 and #385) but I don't think that is a big deal (https://github.com/radiantearth/stac-api-spec/blob/master/CHANGELOG.md#changed).

Also PATCH support is missing: 

- #375

To clarify, 
- #385 
is not a recent change in Txn Ext -- it's always been defined at this path, it's just that the sfapi impl is incorrect. Only 
- #384
- is a change for 1.0.0-rc.1. 

To the overall point, there is an "attempt" to implement Txn Ext, so it's reasonable to advertise it, even if it isn't entirely conformant.

Issue #376, opened 2022-03-16
TransactionsClient should provide option for not returning an Item from endpoints


philvarner:
The Transaction Extension allows for either a 200, 202, or 204 success response from any of the endpoints. 

Currently, all of the TransactionsClient methods have signatures that indicate they are supposed to return an Item.

I would like to see these two changes:

- change the signatures from returning `Item` to returning `Optional[Item]`
- if None is returned, the status code 204 No Content should be returned
- if Item is returned, the status code 200 OK should be returned

PR #387, opened 2022-03-25
change transactions clients method signatures to match use, allow result of None to return 204

philvarner:
**Related Issue(s):** 

- https://github.com/stac-utils/stac-fastapi/issues/376

**Description:**

- for all Txn clients, update the method return types to match what can actually be returned. This is now None, a STAC object type, or a starlette Response.
- In addition to the existing behavior of returning a 200 + STAC object type or Response, also return a 204 No Content if the result value is None.

**PR Checklist:**


moradology:
Discussed this with @geospatial-jeff. On the one hand, it is generally the case that exceptions are used to inject non-200 responses. At the same time, we agreed it felt bizarre to force exceptions in non-exceptional cases. +1

moradology (approved):

Issue #379, opened 2022-03-24
Dockerfile improvements, and and deployable version (comments and questions)


philvarner:
I've made some improvements to the Dockerfile used for development in stac-fastapi-elasticsearch https://github.com/stac-utils/stac-fastapi-elasticsearch/blob/main/Dockerfile.dev

and created one that generates a deployable image https://github.com/stac-utils/stac-fastapi-elasticsearch/blob/main/Dockerfile.deploy

I can port these changes to here if desired.

A few of the concerns I had about the existing Dockerfile are:

I'm not sure why it uses a 2-phase build with no directives in the first phase?

```
FROM python:3.8-slim as base

FROM base as builder
```

I don't think we have any requirements that need building from source or need git:

```
# Any python libraries that require system libraries to be installed will likely
# need the following packages in order to build
RUN apt-get update && apt-get install -y build-essential git
```

This isn't used anywhere in our code, and I couldn't find any tools that require it:

```
ARG install_dev_dependencies=true
```

So it creates an /install dir, but then never uses it?

```
ENV PATH=$PATH:/install/bin

RUN mkdir -p /install && \
```

geospatial-jeff:
Leaving this open for now, I'm hoping that #432 resolves this.

duckontheweb:
This is being tracked for the PgSTAC backend by stac-utils/stac-fastapi-pgstac#10.

PR #391, opened 2022-03-26
refactor Makefile and Dockerfile, and update README with better dev instructions

philvarner:
If these changes are acceptable, we also need to figure out what should be in the Readme vs. Contributing vs. the published docs, and what should be in the root readme vs. the pgstac readme

**Related Issue(s):** 

- https://github.com/stac-utils/stac-fastapi/issues/379

**Description:**

* refactor makefile targets to be more clear and work better for dev
* rework multi-stage Dockerfile build to run faster and remove unused directives
* update README with better instructions for running

**PR Checklist:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,22 @@
 FROM python:3.8-slim as base
 
-FROM base as builder
 # Any python libraries that require system libraries to be installed will likely
 # need the following packages in order to build
-RUN apt-get update && apt-get install -y build-essential git
+RUN apt-get update && \
+    apt-get -y upgrade && \
+    apt-get install -y build-essential git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 ENV CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
 
-ARG install_dev_dependencies=true
+FROM base as builder
 
 WORKDIR /app
 
-# Install stac_fastapi.types
 COPY . /app
 
-ENV PATH=$PATH:/install/bin
-
-RUN mkdir -p /install && \
-    pip install -e ./stac_fastapi/types[dev] && \
+RUN pip install -e ./stac_fastapi/types[dev] && \
     pip install -e ./stac_fastapi/api[dev] && \
     pip install -e ./stac_fastapi/extensions[dev] && \
     pip install -e ./stac_fastapi/sqlalchemy[dev,server] && \
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ test-sqlalchemy: run-joplin-sqlalchemy
 
 .PHONY: test-pgstac
 test-pgstac:
-	$(run_pgstac) /bin/bash -c 'export && ./scripts/wait-for-it.sh database:5432 && cd /app/stac_fastapi/pgstac/tests/ && pytest -vvv'
+	$(run_pgstac) /bin/bash -c 'export && ./scripts/wait-for-it.sh database:5432 && cd /app/stac_fastapi/pgstac/tests/ && pytest -vvv --asyncio-mode=auto'
 
 .PHONY: run-database
 run-database:
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 APP_HOST ?= 0.0.0.0
 APP_PORT ?= 8080
 EXTERNAL_APP_PORT ?= ${APP_PORT}
-run_docker = docker-compose run --rm \
+run_sqlalchemy = docker-compose run --rm \
 				-p ${EXTERNAL_APP_PORT}:${APP_PORT} \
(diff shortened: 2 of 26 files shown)


Dockerfile
```diff
@@ -1,23 +1,22 @@
 FROM python:3.8-slim as base
 
-FROM base as builder
 # Any python libraries that require system libraries to be installed will likely
 # need the following packages in order to build
-RUN apt-get update && apt-get install -y build-essential git
+RUN apt-get update && \
```
philvarner, line 5: this should be in the base stage, since it doesn't need to be re-executed each time and can be cached

Dockerfile
```diff
@@ -1,23 +1,22 @@
...
+    apt-get -y upgrade && \
+    apt-get install -y build-essential git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 ENV CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
 
-ARG install_dev_dependencies=true
```
philvarner, line 10: this doesn't do anything AFAIK

Dockerfile
```diff
@@ -1,23 +1,22 @@
...
+FROM base as builder
 
 WORKDIR /app
 
-# Install stac_fastapi.types
 COPY . /app
 
-ENV PATH=$PATH:/install/bin
```
philvarner, line 17: /install doesn't seem to be used for anything?

Makefile
```diff
@@ -18,29 +18,33 @@ run_pgstac = docker-compose run --rm \
 image:
 	docker-compose build
 
-.PHONY: docker-run
-docker-run: image
-	$(run_docker)
+.PHONY: docker-run-all
```
philvarner, line 21: added this so all the commands can be done with make rather than docker-compose.

Makefile
```diff
@@ -18,29 +18,33 @@ run_pgstac = docker-compose run --rm \
...
+docker-run-all:
+	docker-compose up
+
+.PHONY: docker-run-sqlalchemy
```
philvarner, line 25: explicitly renamed all the targets like `docker-run` to `docker-run-sqlalchemy`, since that's what it does.

Makefile
```diff
@@ -18,29 +18,33 @@ run_pgstac = docker-compose run --rm \
...
 	$(run_pgstac) /bin/bash
 
 .PHONY: test-sqlalchemy
 test-sqlalchemy: run-joplin-sqlalchemy
-	$(run_docker) /bin/bash -c 'export && ./scripts/wait-for-it.sh database:5432 && cd /app/stac_fastapi/sqlalchemy/tests/ && pytest'
+	$(run_sqlalchemy) /bin/bash -c 'export && ./scripts/wait-for-it.sh database:5432 && cd /app/stac_fastapi/sqlalchemy/tests/ && pytest -vvv'
 
 .PHONY: test-pgstac
 test-pgstac:
-	$(run_pgstac) /bin/bash -c 'export && ./scripts/wait-for-it.sh database:5432 && cd /app/stac_fastapi/pgstac/tests/ && pytest -vvv --asyncio-mode=auto'
```
philvarner, line 43: there's another PR where this gets added to pytest.ini, merge that PR first.

stac_fastapi/pgstac/README.md
```diff
@@ -47,35 +46,16 @@ pip install -e \
     stac_fastapi/pgstac[dev,server]
 ```
 
-## Local Development
```
philvarner, line 50: duplicated from root Readme now.

Dockerfile
geospatial-jeff: It's used as a directory to hold the outputs of `pip install -e` but I don't think its needed.  The upside is it makes the WORKDIR a little cleaner, downside is it makes the Dockerfile a little more complex.  Personally I'm fine removing it.

geospatial-jeff (approved):

Dockerfile
geospatial-jeff: Yup, this was used before we switched away from pipenv.

Dockerfile
philvarner: Could you explain to me how it's configured to be used as a directory to hold the pipe install -e outputs?

Makefile
philvarner: actually, went ahead and reverted it.

Dockerfile
philvarner: Ah, I think I actually just stumbled upon the example where this dockerfile came from: https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3

What's missing is that pip isn't being invoked with `--install-option="--prefix=/install"`, so it's not actually being used. (I thought this might be the case, but there's a lot of implicit python magic I don't know about yet)

Issue #357, opened 2022-02-24
In landing page, POST /search link advertises type as application/json instead of application/geo+json


philvarner:
<img width="432" alt="image" src="https://user-images.githubusercontent.com/124996/155593073-fdc2d849-811f-4dc9-82e0-3df8a86fa4d7.png">

NFuture:
I am new to opensource but I have learned python and other languages. How can I contribute

NFuture:
cant find where the stac_pydantic is located at to see the LandingPage which the response model which is declared in the api/app.py file

philvarner:
@NFuture look at the LandingPageMixin class

PR #393, opened 2022-03-28
fix search, next, and prev media types

philvarner:
**Related Issue(s):** 

- https://github.com/stac-utils/stac-fastapi/issues/357

**Description:**

- root POST search link advertises type `application/geojson` (only GET was doing this)
- pagination links advertise `application/geojson`

**PR Checklist:**

### DIFF
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/models/links.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/models/links.py
@@ -125,7 +125,7 @@ class PagingLinks(BaseLinks):
                 href = merge_params(self.url, {"token": f"next:{self.next}"})
                 link = dict(
                     rel=Relations.next.value,
-                    type=MimeTypes.json.value,
+                    type=MimeTypes.geojson.value,
                     method=method,
                     href=href,
                 )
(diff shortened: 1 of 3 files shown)


stac_fastapi/pgstac/stac_fastapi/pgstac/models/links.py
```diff
@@ -149,14 +149,14 @@ def link_prev(self) -> Optional[Dict[str, Any]]:
...
-                    type=MimeTypes.json.value,
+                    type=MimeTypes.geojson.value,
                     method=method,
                     href=href,
                 )
             if method == "POST":
                 return {
                     "rel": Relations.previous,
-                    "type": MimeTypes.json,
+                    "type": MimeTypes.geojson,
```
moradology, line 159: I'm not sure I see why we use `.value` in one case but not the other. Obviously you didn't introduce this asymmetry but I wonder if you got some sense of what the distinction is while working on this portion of the library

philvarner: I'd wondered about that too, but didn't go down that rabbit hole. It looks like with the way MimeTypes is defined (`MimeTypes(str, Enum)`) that these are identical. We should remove the use of `value` from all the MimeTypes.

philvarner:
@moradology I don't have Write permission (I guess that's the one I need?) so I can't merge this PR. I'd be glad to get that permission, since I have a number of other PRs to merge.

moradology:
I'm planning to merge these all, actually. Just want to go ahead and get everything approved so I can think a bit about how to reduce the complexity of the inevitable rebases towards the end of the process... As for giving you perms, I don't think I have the power to modify organization privileges. @lossyrob might be able to help

Issue #396, opened 2022-04-27
PgSTAC: Add support for rehydrating search results within API


mmcfarland:
[PgSTAC 0.5.1](https://github.com/stac-utils/pgstac/releases/tag/v0.5.1) introduced a `nohydrate` runtime option that prevents the database from hydrating items with the collection's base item. This is useful if database CPU is high under load. To distribute the CPU load to API servers and free up resources, the pgstac backend in stac-fastapi should post process results to include base_item properties and also apply the fields extension to the item results.

geospatial-jeff:
Closed with #397

PR #397, opened 2022-04-27
PgSTAC: API hydration of search result items

mmcfarland:
**Related Issue(s):** 

- #396

**Description:**

- Adds support for pgstac backend to hydrate search results with base_items when the `nohydrate` option is set.
- Aligns pgstac backend Fields Extension with pgstac psql implementation

**PR Checklist:**


stac_fastapi/pgstac/setup.py
```diff
@@ -16,6 +16,7 @@
     "asyncpg",
     "buildpg",
     "brotli_asgi",
+    "psycopg[binary]",
```
mmcfarland, line 19: I think this needs to be added to pypgstac dependencies.

lossyrob: https://github.com/stac-utils/pgstac/issues/105

vincentsarago: I don't think we need psycopg in stac-fastapi.pgstac because we are using asyncpg

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
```diff
@@ -0,0 +1,120 @@
...
+                    # Matching key/value is a no-op
+                    pass
+                elif isinstance(b[key], list) and isinstance(i.get(key), list):
+                    # Merge unequal lists
+                    i[key].extend(deepcopy(b[key]))
+            else:
+                # Keys in base item that are not in item are simply copied over
+                i[key] = deepcopy(b[key])
+
+    base_item = await base_item_cache.get(item["collection"])
```
TomAugspurger, line 38: So this would be removed and become a parameter

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
```diff
@@ -0,0 +1,120 @@
...
+            if key in i:
+                if isinstance(b[key], dict) and isinstance(i.get(key), dict):
+                    # Recurse on dicts to merge values
+                    merge(b[key], i[key])
+                elif b[key] == i.get(key):
+                    # Matching key/value is a no-op
+                    pass
+                elif isinstance(b[key], list) and isinstance(i.get(key), list):
+                    # Merge unequal lists
+                    i[key].extend(deepcopy(b[key]))
```
TomAugspurger, line 33: What's the reason for the `deepcopy` rather than something like

```python
i[key].extend(b[key])
```

I suppose then we're going to later mutate `i[key]`, and we don't want to mutate the base `b[key]`?

Can you confirm that we don't have cases where we repeatedly copy the same data? i.e. we do a `deepcopy` on `b[key]`, but then we later do a `deepcopy` on `b[key][0]`, `b[key][1]`, ... as we call `merge` recursively?

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
```diff
@@ -0,0 +1,120 @@
...
+            else:
+                # Keys in base item that are not in item are simply copied over
+                i[key] = deepcopy(b[key])
+
+    base_item = await base_item_cache.get(item["collection"])
+    merge(base_item, item)
+    return Item(**item)
```
TomAugspurger, line 40: Just confirming that we need a `pystac.Item` here, rather than a regular `dict`?

I'm not familiar with stac-fastapi so maybe we already used items in some places. We'll need to be careful with performance here, to make sure no expensive operations are happening in `Item.__init__`.

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
```diff
@@ -0,0 +1,120 @@
+"""stac-fastapi utility methods."""
+from copy import deepcopy
+from typing import Any, Dict, Optional, Set, Union
+
+from stac_fastapi.pgstac.types.base_item_cache import BaseItemCache
+from stac_fastapi.types.stac import Item
+
+
+async def hydrate(
```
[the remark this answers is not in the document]
mmcfarland: Hydrate has been moved outside of this project, but base_item is now passed in and as now a sync function.

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
mmcfarland: That's correct, `i[key]` may be mutated later and `b[key]` may have been applied to many instances of `i`.

I can confirm that at the point of the `deepcopy` (in the new implementation in pypgstac), the code will never recurse deeper into any paths beneath `b[key]` so we shouldn't be copying the same data more than once.

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
mmcfarland: Technically, that [`Item` was a `TypedDict`](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/types/stac_fastapi/types/stac.py#L49) and not a `pystac.Item` -- but yes, I had some concern if even that conversion was an unneccessary cost (though I suspect not much). Regardless, after migrating the function to pypgstac, it operates only on `Dict[str, Any]` and doesn't initialize any other types.

stac_fastapi/pgstac/setup.py
mmcfarland: It's now been moved to pypgstac and only included as a dev dependency here to allow pypgstac to manage the unit test database in conftest.py.

stac_fastapi/pgstac/stac_fastapi/pgstac/utils.py
```diff
@@ -0,0 +1,115 @@
...
+                # The key to remove does not exist on the source, so it is ignored
+                pass
+
+    # Coalesce incoming type to a dict
+    item = dict(item)
+
+    clean_item = include_fields(item, include)
+
+    # If, after including all the specified fields, there are no included properties,
+    # return the full item.
```
mmcfarland, line 92: The spec is ambiguous about behavior here. Likely, someone would have entered a typo or had a mistaken understanding of attributes on the item. Rather than return a completely empty item, this behavior treats it as if they included an empty `include` property, which is to include everything.

mmcfarland: Behavior as implemented is now `{id, collection}`

stac_fastapi/api/stac_fastapi/api/errors.py
```diff
@@ -23,7 +23,7 @@
 DEFAULT_STATUS_CODES = {
     NotFoundError: status.HTTP_404_NOT_FOUND,
     ConflictError: status.HTTP_409_CONFLICT,
-    ForeignKeyError: status.HTTP_422_UNPROCESSABLE_ENTITY,
+    ForeignKeyError: status.HTTP_424_FAILED_DEPENDENCY,
```
mmcfarland, line 26: Just flagging that this change affects all backends.

mmcfarland:
@lossyrob @bitner Got all the tests passing, this is ready for a final review.

stac_fastapi/pgstac/setup.py
```diff
@@ -13,20 +13,22 @@
...
     "asyncpg",
     "buildpg",
     "brotli_asgi",
     "pygeofilter @ git+https://github.com/geopython/pygeofilter@v0.1.1#egg=pygeofilter",
+    "pypgstac==0.6.0",
 ]
 
 extra_reqs = {
     "dev": [
+        "pypgstac[psycopg]==0.6.0",
```
lossyrob, line 26: Seems like we could move this to `install_requires` and remove `psycopg[binary]`, given that `pypgstac[psycopg]` contains pyscopyg[binary] and [pool]

lossyrob: Actually, now I'm confused - aren't we avoiding psycopy in the standard install, and only including it in the dev extras? Why is `psycopg[binary]` required in `install_requires` above?

lossyrob:
@mmcfarland one outstanding question about dependencies, but otherwise LGTM! Will hit approve after that convo wraps up.

stac_fastapi/pgstac/setup.py
mmcfarland: Ah, good catch. `psychopg[binary]` was added back in https://github.com/stac-utils/stac-fastapi/pull/397/commits/efc6e6404bd8920c5dc389f0069b5ee13e167e11#diff-55ad2c09aab406c3bab2128a485d2faaabe4db8c50d3b3c8efc69a9cc8b3c4a1 and was only meant to be temporary until pypgstac was released. I'll remove it.

mmcfarland: Removed in 9a70a02

lossyrob (approved):

Issue #350, opened 2022-02-18
Item bbox field should be nullable in the sqlalchemy backend


yellowcap:
## Problem
I have some trouble with the item `bbox` field when ingesting STAC items that have a `null` value on the geometry field. The STAC spec allows the item bbox to be null, when the geometry field is null, but in the sqlalchemy backend the bbox is not nullable.

## Example STAC Item with a null geometry and no bbox
I came across this problem when trying to ingest STAC Item data from the usgs-landsat stac api.

In this api, there are some items that have `geometry` set to `null` and omit the bbox field. Here is a link to an example:

https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l3-ba/items/LE07_CU_002012_20150101_20210502_02_BA

## Proposed solution
I would like to propose to make the bbox field nullable in the items table. This would increase the compliance with the STAC spec.

geospatial-jeff:
@yellowcap Good catch!  We'll have to start with an upstream change to `stac-pydantic`, I created an issue there!

yellowcap:
Excellent, thanks @geospatial-jeff . I had a go at the `stac-pydantic` change. But I think that we need to go one layer up, as the `geometry` field is defined in  `geojson-pydantic` here: https://github.com/developmentseed/geojson-pydantic/blob/master/geojson_pydantic/features.py#L19

PR #398, opened 2022-04-29
Make item geometry and bbox nullable for sqlalchemy backend

yellowcap:
**Related Issue(s):** 

- #350

**Description:**
Geometry and bbox fields of the Item table need to be nullable to be compliant with the stac standard. This PR makes the two fields nullable for the sqlalchemy backend.

Most upstream fixes in the pydantic classes have already been merged. One bug fix here https://github.com/stac-utils/stac-pydantic/pull/114 is still in review.

The stac-pydantic library needs to have a new release before this PR can be merged, as the updates in the pydantic validator classes need to be installed so they accept null values for these two fields.

The new test of this PR will pass with the next release of stac-pydantic.

**PR Checklist:**


yellowcap:
Ok so the upstream changes are all merged. Since version 2.0.3 of stac-pydantic, the geometry validator includes the nullable option. So the tests pass now and this change is ready for merging from my point of view.

@lossyrob would you mind having a look? Thanks!

lossyrob (approved):

PR #417, opened 2022-06-27
Update filter conformance classes

m-mohr:
The filter conformance classes seem to be completely wrong. They don't even match the (wrong) ones in rc.1.

See also https://github.com/radiantearth/stac-api-spec/pull/312

I'm not sure which one of the constants if for ogcapi-features though, so the changes may need to be swapped.

Looking at this, the question also arises why for example `https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features#sort` is not used anywhere in stac-fastapi?

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
@@ -27,9 +27,7 @@ class FilterConformanceClasses(str, Enum):
     """
 
     FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
-    ITEM_SEARCH_FILTER = (
-        "https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features#filter"
-    )
+    ITEM_SEARCH_FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
     CQL_TEXT = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:cql-text"
     CQL_JSON = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:cql-json"
     BASIC_CQL = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:basic-cql"
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
@@ -23,7 +23,7 @@ class FilterConformanceClasses(str, Enum):
 
     FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
     ITEM_SEARCH_FILTER = (
-        "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
+        "https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features#filter"
     )
     CQL_TEXT = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:cql-text"
     CQL_JSON = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:cql-json"
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
@@ -27,7 +27,9 @@ class FilterConformanceClasses(str, Enum):
     """
 
     FILTER = "http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/filter"
-    FEATURES_FILTER = "http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/features-filter"
+    FEATURES_FILTER = (
+        "http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/features-filter"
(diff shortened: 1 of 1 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
```diff
@@ -21,9 +21,9 @@ class FilterConformanceClasses(str, Enum):
     See https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter
     """
 
-    FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:filter"
+    FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
     ITEM_SEARCH_FILTER = (
-        "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:item-search-filter"
+        "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
```
m-mohr, line 26:
```suggestion
        "https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features#filter"
```

geospatial-jeff:
> Looking at this, the question also arises why for example https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features#sort is not used anywhere in stac-fastapi?

The sort extension currently uses `https://api.stacspec.org/v1.0.0-rc.1/item-search#sort`, @m-mohr should it use the `ogcapi-features` href instead?

stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
```diff
@@ -26,9 +26,9 @@ class FilterConformanceClasses(str, Enum):
     See https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter
     """
 
-    FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:filter"
+    FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
     ITEM_SEARCH_FILTER = (
-        "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:item-search-filter"
+        "https://api.stacspec.org/v1.0.0-rc.1/ogcapi-features#filter"
```
philvarner, line 31: This should be 
`https://api.stacspec.org/v1.0.0-rc.1/item-search#filter`

https://github.com/radiantearth/stac-api-spec/tree/master/fragments/filter

geospatial-jeff: Awesome I can make that change.

stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
```diff
@@ -26,9 +26,9 @@ class FilterConformanceClasses(str, Enum):
     See https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter
     """
 
-    FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter:filter"
+    FILTER = "https://api.stacspec.org/v1.0.0-rc.1/item-search#filter"
```
philvarner, line 29: I think this is supposed to be two separate ones -- Filter defines that filter is being used, and Features Filter binds it to the OAFeat endpoint

```
Filter: http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/filter
Features Filter: http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/features-filter
```

m-mohr:
> The sort extension currently uses `https://api.stacspec.org/v1.0.0-rc.1/item-search#sort` (see [here](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/extensions/stac_fastapi/extensions/core/sort/sort.py#L25-L27)), @m-mohr should it use the `ogcapi-features` href instead?

Depends on where it is supported, I think.
.../item-search#sort if it is supported for /search
.../ogcapi-features#sort if it is supported for /collections/...

philvarner:
> 

Correct -- except for filter, where we re-use the OGC conformance class binding filter to OAFeat

m-mohr:
But there are still separate conformance classes, only that there's no "new" one for STAC, right?

stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
geospatial-jeff: https://github.com/stac-utils/stac-fastapi/pull/417/commits/fe840fb79870c0013e14d9a8c08d01729a4c2a1c

philvarner:
> But there are still separate conformance classes, only that there's no "new" one for STAC, right?

correct

stac_fastapi/extensions/stac_fastapi/extensions/core/filter/filter.py
geospatial-jeff: https://github.com/stac-utils/stac-fastapi/pull/417/commits/40b3ae16741a7044bc87a8597232b24185394fb9

geospatial-jeff:
@philvarner How does it look now?   I updated the item search and filter conformance classes, and added the features filter conformance class which was previously missing.

geospatial-jeff (approved):

Issue #406, opened 2022-05-25
Transaction Extension: purpose of the collection_id path parameter


drnextgis:
What is the purpose of `{collection_id}` path parameter? It seems we can put an arbitrary value there and stac-fastapi will take collection name from the body.

![image](https://user-images.githubusercontent.com/866124/170303431-743f0b4d-bae7-4e81-bd4c-331b8fe7d7ca.png)

Since there is no way to specify `collection_id` via UI it uses `{collection_id}` as it is :man_shrugging:

duckontheweb:
The STAC Spec also requires that a link with a `rel` type of `collection` be present if the `collection` property is present. This means that if you are using something like `pystac.Item.validate` to validate Items prior to insertion you have to add that link, even though it is going to be replaced by the backend anyway.

It seems like the desired behavior here would be:
1. Use the `collection_id` path parameter from the request if there is no `collection` property in the body *or* if that property is present and matches the path parameter
2. Raise an exception if the `collection` property is present and does not match the path parameter.

Issue #384, opened 2022-03-25
Transaction Ext: POST and PUT should validate the Item's id and collection match the path's id and collection


philvarner:
E.g., I could have a body like:

```json
{
  "id": "1",
  "collection": "1",
}
```

that I post to `/collections/2/items` or put to `/collections/1/items/1` `/collections/1/items/2`.  Either of these should return a 400.

PR #424, opened 2022-07-19
Return added Item/Collection in Transactions response

### LINKED ISSUE #423: Transactions endpoint responses include incorrect links
When using `POST /collections` to add a new Collection, the `self` and `root` links in the response are the original links from the submitted Collection instead of the links for the newly created Collection in the API.

**Observed Response**
As an example, add the following self link to the example `joplin` Collection:

```json
{
    "rel": "self",
    "href": "/some/path/to/collection.json"
}
```

The `POST /collections` response includes that same `self` link unchanged:

```console
$ curl -X POST -H "Content-Type: application/json" -d "@tests/data/transactions_api/collection.json" -s http://localhost:8080/collections | jq ".links[] | select(.rel == \"self\")"
{
  "rel": "self",
  "href": "/some/path/to/collection.json"
}
```

The same issue exists with `PUT /collections`, `POST /collections/{collection_id}/items`, and `PUT /collections/{collection_id}/items`. 

**Expected Response**
I expected this to instead return the new `self` link:

```console
$  curl -X POST -H "Content-Type: application/json" -d "@tests/data/transactions_api/collection.json" -s http://localhost:8080/collections | jq ".links[] | select(.rel == \"self\")"
{
  "rel": "self",
  "type": "application/json",
  "href": "http://localhost:8080/collections/joplin"
}
```

**Possible Solution**

This can be easily fixed by adding the `CollectionLinks` logic to the `AsyncBaseTransactionsClient.create_collection` handler (and the equivalent handlers for the other endpoints):

```python
async def create_collection(
    self, collection: stac_types.Collection, **kwargs
) -> Optional[Union[stac_types.Collection, Response]]:
    """Create collection."""
    request = kwargs["request"]
    pool = request.app.state.writepool
    await dbfunc(pool, "create_collection", collection)
    collection["links"] = await CollectionLinks(
        collection_id=collection["id"], request=request
    ).get_links(extra_links=collection.get("links"))

    return stac_types.Collection(**collection)
```

Just a note that this would be really helpful in implementing a solution for stac-utils/pystac-client#241 because we could create a `CollectionClient` directly from the `POST` or `PUT` response rather than having to fetch the Collection again in order to get the correct links.

I can work on a PR for this in the `pgstac` backend.

duckontheweb:
**Related Issue(s):** 

- Closes #423 

**Description:**

Updates `stac_fastapi.pgstac.transactions.TransactionsClient` to return the Item or Collection from the Transactions endpoints as it would appear in a subsequent `GET` call. This means that `self` and `root` links point to the API rather than the original HREFs provided in the request.

I did not see any places where the documentation needed updating, but let me know if I missed anything.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/transactions.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/transactions.py
@@ -14,6 +14,7 @@ from stac_fastapi.extensions.third_party.bulk_transactions import (
 from stac_fastapi.pgstac.db import dbfunc
 from stac_fastapi.types import stac as stac_types
 from stac_fastapi.types.core import AsyncBaseTransactionsClient
+from stac_fastapi.pgstac.models.links import CollectionLinks, ItemLinks
 
 logger = logging.getLogger("uvicorn")
 logger.setLevel(logging.INFO)
@@ -22,7 +23,7 @@ logger.setLevel(logging.INFO)
 @attr.s
 class TransactionsClient(AsyncBaseTransactionsClient):
     """Transactions extension specific CRUD operations."""
-
+    
     async def create_item(
         self, collection_id: str, item: stac_types.Item, **kwargs
     ) -> Optional[Union[stac_types.Item, Response]]:
@@ -37,7 +38,12 @@ class TransactionsClient(AsyncBaseTransactionsClient):
         request = kwargs["request"]
         pool = request.app.state.writepool
         await dbfunc(pool, "create_item", item)
-        return item
+        item["links"] = await ItemLinks(
+            collection_id=item["collection"],
+            item_id=item["id"],
+            request=request,
+        ).get_links(extra_links=item.get("links"))
+        return stac_types.Item(**item)
 
     async def update_item(
         self, collection_id: str, item_id: str, item: stac_types.Item, **kwargs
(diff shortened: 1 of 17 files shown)


stac_fastapi/pgstac/stac_fastapi/pgstac/transactions.py
```diff
@@ -21,15 +22,20 @@
...
     async def create_item(
         self, item: stac_types.Item, **kwargs
     ) -> Optional[Union[stac_types.Item, Response]]:
         """Create item."""
         request = kwargs["request"]
         pool = request.app.state.writepool
         await dbfunc(pool, "create_item", item)
-        return item
+        item["links"] = await ItemLinks(
+            collection_id=item["collection"],
```
duckontheweb, line 34: Just a note that if we make the `collection_id` path parameter available here, as suggested in #406 and #384 then we should update this to get the Collection ID from that parameter.

duckontheweb: > Just a note that if we make the `collection_id` path parameter available here, as suggested in #406 and #384 then we should update this to get the Collection ID from that parameter.

#425 makes this change, so we may want to merge that PR first if it is approved.

geospatial-jeff (approved):

PR #425, opened 2022-07-20
Use collection_id path parameter Items Transactions endpoints

### LINKED ISSUE #384: Transaction Ext: POST and PUT should validate the Item's id and collection match the path's id and collection

### LINKED ISSUE #385: Transactions Extension: PUT endpoint is incorrect, should be /collections/{c_id}/items/{i_id}
Currently, the PUT endpoint is `/collections/{c_id}/items`. However, according to the Txn Ext spec, this is supposed to be `/collections/{c_id}/items/{i_id}`

This may be a bit challenging to implement, as there are no endpoints currently that take both a body and path or query parameters, so `_create_endpoint` may need a bit of refactoring to allow both.

### LINKED ISSUE #406: Transaction Extension: purpose of the collection_id path parameter

### LINKED ISSUE #416: Invalid OpenAPI JSON - path parameter collection_id missing in PUT and POST
I'm trying to use the OpenAPI JSON produced by stac-fastapi to set up the API in Azure API Management. When validating the OpenAPI JSON, I get the following error

```
Declared path parameter "collection_id" needs to be defined within every operation in the path (missing in "put", "post"), or moved to the path-level parameters object
```

This can be fixed by manually editing the OpenAPI JSON and adding:

```
"parameters": [
    {
        "description": "Collection ID",
        "required": true,
        "schema": {
            "title": "Collection Id",
            "type": "string",
            "description": "Collection ID"
        },
        "name": "collection_id",
        "in": "path"
    }
],
```

into the definitions for the PUT and POST requests for `/collections/{collection_id}/items`.

I've tried to work out how to fix this myself, and have tracked it down to the differences between the code in `register_get_item_collection` in `app.py` and `register_create_item` and `register_update_item` in `transaction.py`. The former defines a request model:

```
request_model = create_request_model(
            "ItemCollectionURI",
            base_model=ItemCollectionUri,
            mixins=[get_pagination_model],
        )
```
that specifies the collection_id path parameter. But the latter only specify a request model of `Item` which doesn't have that path parameter.

I've tried creating a new request model of:

```
class ItemWithCollectionPathParam(stac_types.Item):
    collection_id: str = attr.ib(default=Path(..., description="Collection ID"))
```

and using that as the request model as follows:

```
def register_create_item(self):
        """Register create item endpoint (POST /collections/{collection_id}/items)."""
        self.router.add_api_route(
            name="Create Item",
            path="/collections/{collection_id}/items",
            response_model=Item if self.settings.enable_response_models else None,
            response_class=self.response_class,
            response_model_exclude_unset=True,
            response_model_exclude_none=True,
            methods=["POST"],
            endpoint=self._create_endpoint(self.client.create_item, ItemWithCollectionPathParam),
        )
```

But that doesn't seem to fix the OpenAPI JSON.

I'm happy to put in a PR to fix this, if someone can give me some advice on where/how to fix it.

duckontheweb:
**Related Issue(s):** 

- Closes #406 
- Closes #384 
- Closes #416 
- Closes #385 

**Description:**

- Updates the Transactions Extension PUT and POST Item endpoints to take a `collection_id` path parameter and verify that this collection ID matches the Item's `"collection"` property (#406, #384). This also fixes #416.
- Updates the Transactions Extension PUT Item endpoint to take an `item_id` path parameter and verify that this Item ID matches the Item's `"id"` property (#385).
- Updates a number of tests that relied on the old behavior and fixes some unformatted f-strings in the tests

**PR Checklist:**


duckontheweb:
I think I'm fairly close to fixing the `sqlalchemy` backend and tests. I'll push some changes shortly to get the CI passing again.

duckontheweb:
> I think I'm fairly close to fixing the `sqlalchemy` backend and tests. I'll push some changes shortly to get the CI passing again.

This has been fixed and this is ready for review.

geospatial-jeff (approved):

stac_fastapi/pgstac/stac_fastapi/pgstac/transactions.py
```diff
@@ -23,18 +24,38 @@ class TransactionsClient(AsyncBaseTransactionsClient):
...
         await dbfunc(pool, "create_item", item)
         return item
 
     async def update_item(
-        self, item: stac_types.Item, **kwargs
+        self, collection_id: str, item_id: str, item: stac_types.Item, **kwargs
     ) -> Optional[Union[stac_types.Item, Response]]:
         """Update item."""
+        body_collection_id = item.get("collection")
+        if body_collection_id is not None and collection_id != body_collection_id:
```
geospatial-jeff, line 47: `collection` is required field, I don't think `body_collection_id` would ever be `None`.

geospatial-jeff:
From the transactions spec (https://github.com/radiantearth/stac-api-spec/blob/master/ogcapi-features/extensions/transaction/README.md):
- POST requests shall populate the `collection` field in the item from the URI.
- PUT requests should return a 400 status code if `id` or `collection` fields are different from those in the URI.

stac_fastapi/pgstac/stac_fastapi/pgstac/transactions.py
duckontheweb: The `collection` field is technically optional according to the [Item fields](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields) spec:

> The id of the STAC Collection this Item references to (see [collection relation type](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#relation-types)). This field is required if such a relation type is present and is not allowed otherwise.

Personally, I would prefer to be able to POST Items without a `collection` link/property since I know it's going to be autogenerated by the backend. If the `collection` property is required in the body then I need to also generate a Link with a `collection` rel type in order to validate the Item, which seems unnecessary since the Link and property will be overwritten anyway.

duckontheweb:
> From the transactions spec (https://github.com/radiantearth/stac-api-spec/blob/master/ogcapi-features/extensions/transaction/README.md):
> 
> * POST requests shall populate the `collection` field in the item from the URI.
> * PUT requests should return a 400 status code if `id` or `collection` fields are different from those in the URI.

This is the current behavior for this PR, except it returns a 409 status code if the `id` or `collection` fields are different from those in the URI. I can update the response to match this.

duckontheweb:
@geospatial-jeff I fixed the response code, so this is ready for final review and/or merge.

@philvarner Let me know if you have any feedback on this.

geospatial-jeff (approved):
LGTM!

PR #442, opened 2022-08-02
allow user to pass middleware options

### LINKED ISSUE #435: Better middleware injection
The current middleware injection makes it difficult for users to configure the middleware they are injecting, see https://github.com/stac-utils/stac-fastapi/pull/341 and https://github.com/stac-utils/stac-fastapi/pull/341#issuecomment-1201581834.

There is no real advantage to abstracting middleware here, should just let the user configure and attach their own while providing sensible defaults.

geospatial-jeff:
**Related Issue(s):** 

- Closes #435

**Description:**
The goal of this PR is to make it easier for users to configure middleware passed to the application.  Starlette / fastapi expect the middleware class and arguments to that class to be passed separately, the `starlette.middleware.Middleware` class encapsulates this.
- Reworks middleware injection to allow user to pass arguments to each middleware.  This is a breaking change to the type signature of `StacApi.middleware` instance variable (`List[type]` -> `List[Middleware]`)
- Adds the `StacApi.add_middleware` for adding middleware to the underlying fastapi application.  This method uses the same business logic as `Starlette.add_middleware`.

### Before

```python
from starlette.middleware.cors import CORSMiddleware

# No way to specify middleware arguments.
api = StacApi(
    middlewares=[CORSMiddleware]
)

```

### After
```python
from starlette.middleware import Middleware
from starlette.middleware.cors import CORSMiddleware

# Create middleware w/ options.
cors_middleware = Middleware(
    CORSMiddleware,
    allow_origins=("*",),
    allow_methods=("POST", "GET",)
)

# Add in StacApi constructor.
api = StacApi(
    middleware=[cors_middleware]
)

# Or add after-the-fact.
api.add_middleware(cors_middleware)
```

**PR Checklist:**


gadomski:
@geospatial-jeff is this PR still a good change? If so, since it's breaking, we should decide if we want it to go in before or after #450.

gadomski:
After looking at this again, IMO this is a non-breaking change to the type signature of `middlewares`, since the contained type of the list was undefined previously, and the default argument made it pretty clear what that type _should_ be. So IMO this would be ok to include in 2.5.

jonhealy1 (approved):
Really nice work here!

vincentsarago:
Let's hold of til #625 get's merged but this PR is ✅

vincentsarago (approved):

pedro-cf:
Greetings,

How does one convert this type of middleware definition with this new additon?

```python
@app.middleware("http")
async def add_process_time_header(request: Request, call_next):
    start_time = time.time()
    response = await call_next(request)
    process_time = time.time() - start_time
    response.headers["X-Process-Time"] = str(process_time)
    return response
```

pedro-cf:
Greetings @vincentsarago @geospatial-jeff @jonhealy1,

How do I convert this type of middleware definition with these new changes?

```python
@app.middleware("http")
async def add_process_time_header(request: Request, call_next):
    start_time = time.time()
    response = await call_next(request)
    process_time = time.time() - start_time
    response.headers["X-Process-Time"] = str(process_time)
    return response
```

vincentsarago:
@pedro-cf using `http` middleware is deprecated (or at least not encouraged by starlette). you can easily create a pure ASGI middleware like https://github.com/developmentseed/titiler/blob/main/src/titiler/core/titiler/core/middleware.py#L66-L100 which will be easier to use in stac-fastapi

Issue #432, opened 2022-07-27
Split backends into independent repos


geospatial-jeff:
Discussion started here: https://github.com/stac-utils/stac-fastapi/issues/409#issuecomment-1192699464.

Including backends in this repo was convenient when stac-fastapi was first created but the pros are now outweighed by the cons.

Migration plan:
1. Merge all open PRs.
2. Release a new version, this will be the last version with a monorepo.
3. Create two new repos in `stac-utils` org, one for each backend.
4. Copy backend submodules over to new repos.
5. Merge the `api`, `extensions`, and `types` submodules into a single stac-fastapi package, keep import paths the same to avoid breaking.
6. Cut over to the new stac-fastapi package in the backend repos.
7. Test / validate backend repos before we merge all of the code.
8. Move backend specific issues to the appropriate repo.

duckontheweb:
I'm a huge +1 for this and I'm happy to help with part of it if you want.

> 5\. Merge the `api`, `extensions`, and `types` submodules into a single stac-fastapi package, keep import paths the same to avoid breaking.

I'm a big fan of this. I've been using `poetry` a bunch lately, but it doesn't support installing from GitHub subdirectories, which has been a blocker for trying to install the latest unstable version of this library. This would solve that issue.

geospatial-jeff:
> I'm a huge +1 for this and I'm happy to help with part of it if you want.

That would be great!  I'm going to start working getting all of the open PRs resolved, not exactly sure how long that will take.

duckontheweb:
Sounds good. I can create new repos for the backends (`stac-fastapi-sqlalchemy` and `stac-fastapi-pgstac`). We will probably want to resolve the following PRs before we port the code over since they all make changes to one or both of the backends:

- #425 
- #424 
- #418

geospatial-jeff:
@duckontheweb I can start moving code into `stac-fastapi-sqlalchemy`

duckontheweb:
> @duckontheweb I can start moving code into `stac-fastapi-sqlalchemy`. I'll try to keep the repo structure as similar as possible.

Sounds good. I’m wondering if we should actually fetch from this repo into those ones in order to preserve the Git history. I think setting up this repo as a remote and then doing a “git reset —hard” to the master branch of this repo would do the trick

duckontheweb:
[...]

PR #453, opened 2022-08-11
Improve separation of concerns between core and backend libraries

duckontheweb:
**Related Issue(s):** 

- [#432(comment)](https://github.com/stac-utils/stac-fastapi/issues/432#issuecomment-1210024475)

**Description:**

Demonstrates one approach for creating a clearer separation of concerns between the core `stac-fastapi` package and the backend packages like `stac-fastapi.pgstac`. This draft targets the landing page and collections endpoints to demonstrate the approach; if we decide to go this route, we should apply this to all other route handlers as well.

The general approach is as follows:

- Any methods on the `*BaseCoreClient` classes that are used directly as route handlers have been renamed with a `handle_` prefix. These will be concrete methods defined in this library responsible for parsing the request, calling the appropriate backend method(s), and constructing the response.
- New abstract methods have been created that must be implemented in the backend packages. These methods are all prefixed with `fetch_` and are responsible for fetching the appropriate objects from the database or other backend. These methods have well-defined, typed inputs rather than accepting arbitrary `**kwargs`.

This approach should make it easier to handle all of the logic around constructing links, conformance classes, and other API-layer concerns in this package, ensuring that responses are consistent across the different backends.

Some future TODOs:

- Having more consistent type annotations and type checking in this package would make it easier to ensure that backend applications are implementing methods correctly. This seems out of scope for this PR, but I can start implementing this once this and #450 have been resolved.
- We may want to go one step further and have the `fetch_` and `handle_` methods in separate classes or mixins. My preference for this draft was to focus on the functional changes and leave that kind of structural refactor to a later PR, but I can always experiment with changing the class structure if we decide we want that right away.

I will open a PR in the `stac-fastapi-pgstac` repo to demonstrate the corresponding changes in that backend.

**EDIT:** Note that this PR is against the `remove/backend-packages` branch instead of `master` to simplify the changes. Once #450 has been merged we can update this to use `master` as the base.

**PR Checklist:**

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+tests
 docker-compose.yml
 **/__pycache__
 *.pyc
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,9 +9,9 @@ services:
       dockerfile: Dockerfile
     platform: linux/amd64
     volumes:
-      - ./stac_fastapi:/app/stac_fastapi
-      - ./scripts:/app/scripts
-      - ./tests:/app/tests
+      - ./stac_fastapi:/opt/src/stac_fastapi
+      - ./scripts:/opt/src/scripts
+      - ./tests:/opt/src/tests
     command:
       - bash
       - -c
--- a/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/app.py
@@ -205,7 +205,9 @@ class StacApi:
             response_model_exclude_none=True,
             methods=["POST"],
             endpoint=self._create_endpoint(
-                self.client.handle_post_search, self.search_post_request_model, GeoJSONResponse
+                self.client.handle_post_search,
+                self.search_post_request_model,
+                GeoJSONResponse,
             ),
         )
 
@@ -227,7 +229,9 @@ class StacApi:
             response_model_exclude_none=True,
             methods=["GET"],
             endpoint=self._create_endpoint(
-                self.client.handle_get_search, self.search_get_request_model, GeoJSONResponse
+                self.client.handle_get_search,
+                self.search_get_request_model,
+                GeoJSONResponse,
             ),
         )
 
--- a/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/app.py
@@ -143,7 +143,7 @@ class StacApi:
             response_model_exclude_none=True,
             methods=["GET"],
             endpoint=self._create_endpoint(
-                self.client.landing_page, EmptyRequest, self.response_class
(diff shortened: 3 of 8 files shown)


.dockerignore
```diff
@@ -1,3 +1,4 @@
+tests
```
duckontheweb, line 1: This change is again not directly related to this refactor, but helped speed up the Docker builds

geospatial-jeff:
I like the approach of separating link injection from backend specific logic, my main concern here is breaking the client interfaces and adding extra boilerplate to the clients.  I was thinking that instead of adding more methods we could handle this with a function in the API layer that is called in the FastAPI endpoint after each client method (ex. `client.get_collection`).  The function signature could look something like:

```python
import typing

from stac_fastapi.types.stac import Collection, Collections, Item, ItemCollection

StacTypes = typing.Union[Collection, Collections, Item, ItemCollection]

def hydrate_links(request: Request, data: StacTypes) -> StacTypes:
    ...

```

The `request` object has the URL of the request which lets us infer parent/child/root links.  And we could infer the type of `data` using something equivalent to [`pystac.serialization.identity.identify_stac_object_type`](https://github.com/stac-utils/pystac/blob/69fcf3f2137045bd45ac5cf6cbd2fdad869eb169/pystac/serialization/identify.py#L179-L236).

The [route handler(s)](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/stac_fastapi/api/routes.py) would then look something like this:

```python

def _endpoint(
    request: Request,
    request_data: request_model
):
    # Request data from the backend.
    backend_response = func(request_data, request=request)

    # Hydrate links
    response_with_links = hydrate_links(request, backend_response)

    # Serialize response
    return _wrap_response(response_with_links, response_class)
```

geospatial-jeff:
Circling back to this after thinking for a while, I think the most important thing for me is to remove the need to inject the starlette `Request` object into the backend clients as that is a clear as that violates separation of concerns.  The backend clients should just take request parameters for the given endpoint and fetch data that meets those requirements, and shouldn't be concerned with things like the base url of the request.

duckontheweb:
Thanks @geospatial-jeff, all of your points are right on. I'll update this PR to try to take more of this approach.

> my main concern here is breaking the client interfaces and adding extra boilerplate to the clients.

Yeah, I agree. I think the approach you outlined above would still break the client interfaces since they rely on the `request` being passed as a keyword arg. However, it minimizes the changes to the overall interface, and puts the logic for handling links in once place, which I like.

gadomski:
Closing as OBE #555.

PR #472, opened 2022-10-03
[WIP] Generate links in api layer

geospatial-jeff:
**Related Issue(s):** 

- #432 
- Related to https://github.com/stac-utils/stac-fastapi/pull/453

**Description:**

Moves link generation into the API layer to create better separation of concerns between API/backend libraries.  Links are split into two categories:
1. Inferred links (`self`, `root`, `parent`, `child` etc.) are dynamically generated by the API layer based on the request path and the stac object returned by that request.
2. Backend specific links (`next`, `prev`, links to external resources etc.) may be returned by backend implementations and resolved by the API layer - `urljoin(base_url, link['href'])`.   The current landing page is a good example of this; links to children collections are generated by the backend (`Backend.all_collections`) and resolved by the API layer.  This allows backends to return their own links, and makes it easy for users of those backends to extend this behavior.

With this PR the starlette request object is no longer passed to the backend client as a key word argument and `core.py` no longer imports HTTP specific things like `urllib.parse.urljoin`, `fastapi.Request`, and `stac_fastapi.types.request.get_base_url`.

**TODOs**
- [ ] Check the API spec to make sure we are returning recommended links for each endpoint.
- [ ] Make sure media types are correct.
- [x] Figure out a good way to differentiate between the two types of links mentioned above.  The API layer should override any inferred links returned by the backend and resolve + pass through any other links.
- [ ] Write tests.

**PR Checklist:**

- [ ] Code is formatted and linted (run `pre-commit run --all-files`)


geospatial-jeff:
A few ideas for differentiating between links:
- Reserve a handful of link relations to be used for inferred links.  Some link relations like `self` are obviously inferred but there is some gray area.  For example the `data` relation is currently used by the landing page for `/collections` but `data` could easily apply to other links too.
- Require that backend implementations return additional link members indicating if the link should be inferred.  The API layer would strip this additional link member from the payload before returning the API response to the caller.  For example:

```json
{
    "rel": "next",
    "type": "application/geo+json",
    "method": "GET",
    "href": "/search?token=12345",
    "inferred": false,
}
```

bitner:
I definitely agree with the third option using precedence.

The other thought I had related to ability to control extensions that add additional links is to use the request.scope['routes'] and request.scope['route'] to introspect the App (same as how the OpenAPI docs are created - and there are some examples there that we could use) and to include any links that are at the next level up as the "self" link. This would mean that when requesting "/" endpoint, that it should auto-discover the "/conformance" endpoint and that it would be able to scan the routes and if the FiltersExtension was enabled, it would also find the "/queryables" endpoint. The mime_type would be able to be found on the response_class.

geospatial-jeff:
> use the request.scope['routes'] and request.scope['route'] to introspect the App

I think this is a really good idea, I can play around with it in another PR.

gadomski:
Closing as OBE #555.

PR #474, opened 2022-10-03
Pgstac queryables

### LINKED ISSUE #469: Queryables link relation and endpoint issues
- [Item Search Filter Ext] / : 'http://www.opengis.net/def/rel/ogc/1.0/queryables' (Queryables) link relation missing from both root and each collection
- [Item Search - Filter Ext] Queryables endpoints return Content-Type header 'application/json', must return 'application/schema+json'
- [Item Search - Filter Ext] Queryables endpoint json schmea'$id' value invalid, must be same as queryables url (instead it's the example value given in the Filter Extension)

bitner:
**Related Issue(s):** 

- Partially fixes #469. This does not add the queryables link to / or /collections/{collection_id}/queryables to keep the scope of this fix to the pgstac backend. Currently there is no (at least that I am aware of) mechanism to inject extra links as part of an extension, so hardcoding the queryables links will result in bad links when the FilterExtension is not enabled. 

**Description:**
Adds support to pgstac backend queryables functionality added in pgstac v0.6.8.
  - Adds /queryables endpoint
  - Adds /collections/{collection_id}/queryables endpoint

**PR Checklist:**


geospatial-jeff:
> Currently there is no (at least that I am aware of) mechanism to inject extra links as part of an extension, so hardcoding the queryables links will result in bad links when the FilterExtension is not enabled.

You may currently do this with conformance classes but not links.  This will be easier to add once https://github.com/stac-utils/stac-fastapi/issues/432 is resolved.  https://github.com/stac-utils/stac-fastapi/pull/472 is part of that larger effort and moves in the direction of making links easier to customize but I hadn't thought of the case where an extension requires additional links so I created #475 to track that.

mmcfarland (approved):

geospatial-jeff (approved):
LGTM 👍

carderne:
Sorry for possibly misplaced question but:
Currently setting a `bbox` in the Collection Filter dialog in stac-browser doesn't work with stac-fastapi - it just calls `/collections{collection_id}/items` with `bbox=1,2,3,4`, which is ignored in the route.

Will this fix that? If not where should I go looking to get that to work? (Aside from monkey-patching in some code to use `bbox` in that route...)

bitner:
> Sorry for possibly misplaced question but: Currently setting a `bbox` in the Collection Filter dialog in stac-browser doesn't work with stac-fastapi - it just calls `/collections{collection_id}/items` with `bbox=1,2,3,4`, which is ignored in the route.
> 
> Will this fix that? If not where should I go looking to get that to work? (Aside from monkey-patching in some code to use `bbox` in that route...)

No, this does not add that capability. As far as i know, that is not part of the stac specification. You can bring things like this up by adding an issue rather than replying to unrelated pull requests.

carderne:
👍 #476

Issue #476, opened 2022-10-06
Collection bbox filtering from stac-browser


carderne:
The collection-level filter in stac-browser makes a request that looks like:
```
/collections{collection_id}/items?bbox=1,2,3,4
```

But the `bbox` query param is ignore in that route in stac-fastapi.pgstac. Eg [this](https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items?limit=1&bbox=25,-30,28,-28) request to the MS PC API, which returns an Item nowhere near the requested bbox.

The route called doesn't do anything with the `bbox`:
https://github.com/stac-utils/stac-fastapi/blob/fbdd9931721a73fbbf54646f853ef720dfb25268/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py#L270-L273

As far as I can tell this is not an issue on stac-browser's side. Would appreciate some help understanding how this is supposed to work, because it seems like there's more going on that just adding `bbox` to the queries passed to `_search_base`? If not I'm happy to work on a PR to do that...

geospatial-jeff:
Same as #380 

The `bbox` and `datetime` query parameters on the `/items` endpoint are part of OGC API - Features and not currently supported by stac-fastapi.  It would be really nice to support them though because this has caused some confusion (as mentioned in #380 ).  PRs are more than welcome! 😄 

In order to implement this:
- Update the abstract base classes ([here](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/types/stac_fastapi/types/core.py#L492) and [here](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/types/stac_fastapi/types/core.py#L687)) to accept optional `bbox` and `datetime` arguments.  
- Update the backend implementations  to implement the two new parameters.
    - The sqlalchemy backend will support this.
    - This may require a change to pgstac, @bitner @vincentsarago would know more.

You can find more information about how `bbox` and `datetime` should be implemented here:
- The STAC OpenAPI fragments - https://github.com/radiantearth/stac-api-spec/blob/main/ogcapi-features/openapi.yaml#L119-L144
- The OGC API - Features specification - https://docs.opengeospatial.org/is/17-069r4/17-069r4.html#_features_rootcollectionscollectioniditems

carderne:
Thanks I didn’t find that duplicate issue somehow. Will dig into this tomorrow.

Issue #380, opened 2022-03-24
/collections/{collectionId}/items endpoint only does not apply bbox or datetime parameters to results


philvarner:
It looks like of the `limit` parameter is the only parameter that gets applied to the `/collections/{collectionId}/items` endpoint. 

This would **probably** be acceptable if the server only advertised the **STAC API - Features** conformance class (https://api.stacspec.org/v1.0.0-beta.4/ogcapi-features"), as the spec says (https://github.com/radiantearth/stac-api-spec/tree/master/ogcapi-features):

```
Implementations may optionally provide support for the full superset of STAC API query parameters to the /collections/{collectionId}/items endpoint, where the collection ID in the path is equivalent to providing that single value in the collections query parameter in STAC API.
```

(It's a separate issue as to if this should require the same as OAFeat, but that's TBD)

However, the server advertises the OAFeat class `http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core`, which means that 
`bbox` and `datetime` must also be accepted as filtering parameters.

m-mohr:
Support for bbox and datetime would be a nice addition, indeed.

carderne:
@philvarner @m-mohr might be worth having your input on this PR: #477

Mec-iS:
can this be closed?

vincentsarago:
yes, we can now extend the `/items` endpoint with bbox, datetime and other parameters

PR #477, opened 2022-10-07
Add bbox and datetime to /collections/{collectionId}/items

carderne:
**Related Issue(s):** 
- #476
- #380 

**Description:**
Add `bbox` and `datetime` query parameters to the `/collections{collection_id}/items` route.
Updated base classes, request models and pgstac and sqlalchemy clients.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/models.py
+++ b/stac_fastapi/api/stac_fastapi/api/models.py
@@ -13,6 +13,7 @@ from stac_fastapi.types.search import (
     APIRequest,
     BaseSearchGetRequest,
     BaseSearchPostRequest,
+    str2list,
 )
 
 
@@ -123,6 +124,8 @@ class ItemCollectionUri(CollectionUri):
     """Get item collection."""
 
     limit: int = attr.ib(default=10)
+    bbox: Optional[str] = attr.ib(default=None, converter=str2list)
+    datetime: Optional[str] = attr.ib(default=None)
 
 
 class POSTTokenPagination(BaseModel):
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 * Add support in pgstac backend for /queryables and /collections/{collection_id}/queryables endpoints with functions exposed in pgstac 0.6.8
 * Update pgstac requirement to 0.6.8
+* Add `bbox` and `datetime` query parameters to `/collections/{collection_id}/items`. [476](https://github.com/stac-utils/stac-fastapi/issues/476) [380](https://github.com/stac-utils/stac-fastapi/issues/380)
 
 ### Changed
 
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/core.py
@@ -248,6 +248,8 @@ class CoreCrudClient(AsyncBaseCoreClient):
     async def item_collection(
         self,
         collection_id: str,
+        bbox: Optional[List[NumType]] = None,
+        datetime: Optional[Union[str, datetime]] = None,
         limit: Optional[int] = None,
         token: str = None,
         **kwargs,
@@ -267,8 +269,21 @@ class CoreCrudClient(AsyncBaseCoreClient):
         # If collection does not exist, NotFoundError wil be raised
         await self.get_collection(collection_id, **kwargs)
 
+        base_args = {
+            "collections": [collection_id],
+            "bbox": bbox,
+            "datetime": datetime,
+            "limit": limit,
+            "token": token,
+        }
+
+        clean = {}
+        for k, v in base_args.items():
+            if v is not None and v != []:
+                clean[k] = v
+
         req = self.post_request_model(
-            collections=[collection_id], limit=limit, token=token
+            **clean,
         )
         item_collection = await self._search_base(req, **kwargs)
         links = await CollectionLinks(
--- a/stac_fastapi/pgstac/tests/api/test_api.py
+++ b/stac_fastapi/pgstac/tests/api/test_api.py
@@ -415,3 +415,51 @@ async def test_collection_queryables(load_test_data, app_client, load_test_colle
     assert "properties" in q
     assert "id" in q["properties"]
     assert "eo:cloud_cover" in q["properties"]
+
+
+@pytest.mark.asyncio
+async def test_item_collection_filter_bbox(
+    load_test_data, app_client, load_test_collection
+):
+    coll = load_test_collection
+    first_item = load_test_data("test_item.json")
+    resp = await app_client.post(f"/collections/{coll.id}/items", json=first_item)
+    assert resp.status_code == 200
+
+    bbox = "100,-50,170,-20"
+    resp = await app_client.get(f"/collections/{coll.id}/items", params={"bbox": bbox})
+    assert resp.status_code == 200
+    resp_json = resp.json()
+    assert len(resp_json["features"]) == 1
+
+    bbox = "1,2,3,4"
+    resp = await app_client.get(f"/collections/{coll.id}/items", params={"bbox": bbox})
+    assert resp.status_code == 200
+    resp_json = resp.json()
+    assert len(resp_json["features"]) == 0
+
+
+@pytest.mark.asyncio
+async def test_item_collection_filter_datetime(
+    load_test_data, app_client, load_test_collection
+):
+    coll = load_test_collection
+    first_item = load_test_data("test_item.json")
+    resp = await app_client.post(f"/collections/{coll.id}/items", json=first_item)
+    assert resp.status_code == 200
+
+    datetime_range = "2020-01-01T00:00:00.00Z/.."
+    resp = await app_client.get(
+        f"/collections/{coll.id}/items", params={"datetime": datetime_range}
+    )
+    assert resp.status_code == 200
+    resp_json = resp.json()
+    assert len(resp_json["features"]) == 1
+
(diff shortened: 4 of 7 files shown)


carderne:
Just getting started on this.

I did `pgstac` by copying some code from the search route, and did `sqlalchemy` by just calling the `get_search` method.

Haven't done any testing or added tests, docs yet.

carderne:
Some comments:
1. Updated the `pgstac` code to use the base_args dict approach from the search route. Bit of duplication but it seems okay.
2. For `sqlalchemy`, I did the same thing with the filtering code from the search route. Could be worth pulling some of that filtering code out into two separate functions, as it's pretty self-contained?
3. Haven't added any tests, but could add some similar to the `search` tests, just checking `bbox` and `datetime` functionality?

Then three separate points that probably need attention:
1. There's some type inconsistency in `datetime` parameters. In the type defs it's `str` but in the clients, it's `Union[str, datetime]`. This causes a type issue [here](https://github.com/stac-utils/stac-fastapi/blob/fbdd9931721a73fbbf54646f853ef720dfb25268/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py#L348) calling `.split()` on it. It should probably just be `str` everywhere, as it's supposed to allow `/`-separated ranges. Could also type it more cleverly and check before `split`ing.
2. `datetime` parameters shadow `from datetime import datetime` eg [here](https://github.com/stac-utils/stac-fastapi/blob/fbdd9931721a73fbbf54646f853ef720dfb25268/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py#L191). Should probably rename to `datetime_range` or something.
3. `bbox` is alternately types as `List[NumType]`, `str` and `BBox`. Should probably be `BBox` everywhere, and the `str2list` converter could be a bit more aggressive in converting it into a `BBox` (a tuple of floats, not just a list of anything). For now I did [this](https://github.com/stac-utils/stac-fastapi/pull/477/commits/5aa95bd1d67fd717d1a6ca022ffa853fd019078a) but not great.

stac_fastapi/api/stac_fastapi/api/models.py
```diff
@@ -123,6 +124,8 @@ class ItemCollectionUri(CollectionUri):
     """Get item collection."""
 
     limit: int = attr.ib(default=10)
+    bbox: Optional[str] = attr.ib(default=None, converter=str2list)
```
m-mohr, line 127: Wouldn't it make sense to convert this directly to e.g. `Optional[List[float]]` for bbox and `Optional[List[Union[datetime, None]]]` for datetime so that you don't need to handle the conversion issues downstream?

carderne: Yes I mentioned that in my comment above point 3.

I held off on changing it until I get some feedback, as that is the current behavior in the similar routes.

carderne: `datetime` is a little more complicated, because it also supports ranges like `2020-01-01/2022-01-01` or `2020-01-01/..`, so it must allow `str` input (or a clever pydantic type enforcing either datetime or something with a slash in it).

m-mohr: Ah, I did not see that comment. Still, a list of datetime/None should work. None could be used for open date ranges as specified also for STAC extents in collections...

carderne: Oh sorry I misunderstood you. Yes that makes sense, but as I've looked into this I've found some inconsistency.

For `sqlalchemy`, I [added code](https://github.com/stac-utils/stac-fastapi/pull/477/commits/a68abdfd2085681df2923402f0fe779c7d21831c) based on the [post_search](https://github.com/carderne/stac-fastapi/blob/3ea9232a5c741cb0389410a18e0deed954be1203/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py#L385-L399) implementation: it has some complicated logic on deciding how to treat the `datetime` parameter, and _treats a single datetime as an exact datetime_. It's also more lenient on datetime formatting.

For `pgstac`, I [just used](https://github.com/stac-utils/stac-fastapi/pull/477/commits/26cd5f3bebd08aa1f8f7bdb0c9c144848d2636eb) `post_request_model`, which [treats a single value as an end-date](https://github.com/carderne/stac-fastapi/blob/3ea9232a5c741cb0389410a18e0deed954be1203/stac_fastapi/types/stac_fastapi/types/search.py#L157-L158), and is stricter on enforcing RFC3339.

So a few questions:
1. Should we update the `sqlalchemy` `post_search` and `item_collection` routes to use `post_request_model`?
2. Should a single datetime be considered as an end date? It seems much more useful (if less obvious) than the unlikely case you're able to hit on the exact datetime of an item...
3. If the latter, datetime could be typed as (being pedantic): `None | tuple[datetime, datetime] | tuple[datetime, None] | tuple[None, datetime]`. This is all already validated in `post_request_model` (see the last link above). I think actually enforcing this typing would break the pgstac implementation, as it passes the raw `datetime/datetime` string to the DB's `search` function.

geospatial-jeff:
Sorry its taken me a while to get around to this.

> There's some type inconsistency in datetime parameters. In the type defs it's str but in the clients, it's Union[str, datetime]. This causes a type issue [here](https://github.com/stac-utils/stac-fastapi/blob/fbdd9931721a73fbbf54646f853ef720dfb25268/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py#L348) calling .split() on it. It should probably just be str everywhere, as it's supposed to allow /-separated ranges. Could also type it more cleverly and check before spliting.

I agree that the `Union[str, datetime]` is messy.  I think `str` is more ambiguous because the spec allows both single dates and date time ranges.  If we were to make changes to datetime handling I would prefer for the API layer to coerce the datetime parameters to a tuple of datetime objects.  Maybe `(None, datetime.datetime`), `(datetime.datetime, None`) for single tailed date ranges and (`datetime.datetime, datetime.datetime`) for two tailed date ranges.

> datetime parameters shadow from datetime import datetime eg [here](https://github.com/stac-utils/stac-fastapi/blob/fbdd9931721a73fbbf54646f853ef720dfb25268/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py#L191). Should probably rename to datetime_range or something.

Parameters aim to use the same language as what is in the API spec for clarity.  I'm fine changing this but I don't think its too straight forward.  Pydantic [does support aliasing](https://pydantic-docs.helpmanual.io/usage/model_config/#alias-precedence) but I don't believe attrs does.

> bbox is alternately types as List[NumType], str and BBox. Should probably be BBox everywhere, and the str2list converter could be a bit more aggressive in converting it into a BBox (a tuple of floats, not just a list of anything). For now I did [this](https://github.com/stac-utils/stac-fastapi/pull/477/commits/82817441a81bbed7c81ec899c994ded6eaa766dc) but it's not great.

I agree with all of this, `bbox` should be consistently typed as `BBox` and `str2list` should be more aggressive.

stac_fastapi/api/stac_fastapi/api/models.py
geospatial-jeff: > Should we update the sqlalchemy post_search and item_collection routes to use post_request_model?

I think the sqlalchemy backend would be better if, like pgstac, it had one function (`_search_base`) which contains most of the important query logic and is wrapped by all of the other methods.

> Should a single datetime be considered as an end date? It seems much more useful (if less obvious) than the unlikely case you're able to hit on the exact datetime of an item...

My interpretation of the [OAF datetime parameter](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameter_datetime) is a single date time is an exact date time (`==`) while date time ranges are either `../end_date` or `start_date/..`.  I agree that `==` is arguably less useful when searching than a datetime range but the distinction is there to let users do both.

> If the latter, datetime could be typed as (being pedantic): None | tuple[datetime, datetime] | tuple[datetime, None] | tuple[None, datetime]. This is all already validated in post_request_model (see the last link above). I think actually enforcing this typing would break the pgstac implementation, as it passes the raw datetime/datetime string to the DB's search function.

Lol I should have read this thread first, I proposed the same thing in PR comment below.  We shouldn't be too concerned about breaking backend implementations w/ changes to the API layer as long as those changes make the API layer stronger and improve separation of concerns (I think here they do).  Backend implementations shouldn't have to guess what information is included in a request, that information should be presented in a way that is clear and easy to use and I'm not in love with the current solution for datetime.  Pgstac can always coerce these back to a string.

geospatial-jeff (approved):
This LGTM, thanks for the PR!  We can continue discussion in this PR, but if its okay with you @carderne I think any other changes should become new issues / new PRs.

carderne:
Sounds good!

To sort-of summarise the other issues:
1. Coerce `datetime` to a well-typed tuple at the API surface, fix in the backends.
2. Same for `bbox`.
3. Update sqlalchemy backend to use a single search route.
4. (Maybe?) fix `datetime` shadowing. Easiest way would be to `from datetime import datetime as dt_lib` (rather than changing the param). As it is now, [this if statement](https://github.com/stac-utils/stac-fastapi/blob/fbdd9931721a73fbbf54646f853ef720dfb25268/stac_fastapi/sqlalchemy/stac_fastapi/sqlalchemy/core.py#L210-L211) (and probably others) is broken.

carderne:
@geospatial-jeff I don’t think I can merge this

carderne:
Fixed conflicts introduced by #482.

carderne:
Fixed the linting issue from the merge above

carderne:
@geospatial-jeff @bitner

Just bumping, any reason not to merge?

Issue #492, opened 2022-11-14
How to disable the `Transaction` extension in the `pgstac` app?


jtherrmann:
We're currently using the `pgstac` app provided [here](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/pgstac/stac_fastapi/pgstac/app.py), but we'd like to disable the Transaction extension. We're reluctant to do all of the configuration of the app from scratch. Is there a way to simply disable the extension in the app after it's already been created?

geospatial-jeff:
Hello @jtherrmann the only way of doing this currently is to configure the app from scratch, the library is designed to do this and we encourage users to configure their own instead of relying on the existing applications (they are really just examples).

That being said we could introduce an environment variable to the pgstac backend which optionally enables or disables the transactions extension, allowing the user to set this at deploy time.  This seems reasonable to me.

jtherrmann:
Thanks for the reply @geospatial-jeff. We like the idea of using an environment variable, so I may open a PR to make that change, if that's alright?

jtherrmann:
I opened https://github.com/stac-utils/stac-fastapi/pull/495 just to demonstrate the approach I was thinking of.

PR #487, opened 2022-11-02
Bumpallowedpygeofilter

bitner:
Bump allowed version of pygeofilter to include 0.2.*.

PR #495, opened 2022-11-16
Add environment variable to disable Transaction extension in provided `pgstac` app

jtherrmann:
See https://github.com/stac-utils/stac-fastapi/issues/492

### DIFF
--- a/stac_fastapi/pgstac/stac_fastapi/pgstac/app.py
+++ b/stac_fastapi/pgstac/stac_fastapi/pgstac/app.py
@@ -1,4 +1,10 @@
-"""FastAPI application using PGStac."""
+"""FastAPI application using PGStac.
+
+Disables the Transaction extension if the environment variable
+DISABLE_TRANSACTION_EXTENSION is set to '1' or 'true'."""
+
+import os
+
 from fastapi.responses import ORJSONResponse
 
 from stac_fastapi.api.app import StacApi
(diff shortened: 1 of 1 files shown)


geospatial-jeff:
@jtherrmann thinking of how to make this more applicable to the other extensions.  How about an environment variable `ENABLED_EXTENSIONS` which accepts a comma-delimited string (ex. `transactions,sort,query`) determining what extensions are added to the application.

If it is not present all extensions would be enabled.

jtherrmann:
@geospatial-jeff Is there a use-case for disabling other extensions? The Transaction extension is the only one that I would want to disable, for security reasons, but I can't think of a reason for disabling any of the others.

gadomski:
CI is breaking because we need to ceil the fastapi version (e.g. cb76e2e25e2fe7c285299bbc690133815432fd9f). @geospatial-jeff if you could approve #487, we can merge that to pick up the fix.

geospatial-jeff (approved):
This LGTM!  Thanks very much @jtherrmann 🙏

jtherrmann:
Thanks @geospatial-jeff! When can we expect a release so that we can use this functionality (without installing from our fork)?

geospatial-jeff:
I can cut the next release this weekend, there are a few other outstanding PRs I want to get merged.

jtherrmann:
@geospatial-jeff Great, thanks!

geospatial-jeff:
Sorry for the delay, I got a nasty case of the flu and have been out the past week.  I will cut a new release today @jtherrmann .  Thanks for your patience!

jtherrmann:
@geospatial-jeff No worries, hope you're feeling better! Thanks!!

PR #498, opened 2022-11-25
Add 3.11 to test matrix

tonybaloney:
The package manifests specify support for any version 3.8>= but 3.11 is missing from the test matrix.

This will run the tests on Python 3.11


geospatial-jeff:
@tonybaloney Thanks for the PR!  I've cherry picked your commit to #500 so I can debug why CI isn't passing.

Issue #454, opened 2022-08-19
OpenAPI servers not auto-generated when root_path is set


duckontheweb:
As of tiangolo/fastapi#1596 FastAPI will autogenerate the `servers` section of the OpenAPI document if the `FastAPI` instance has a `root_path` set. This is crucial in setting up the right routes in things like the "Try it out" section of the Swagger docs page.
Because `stac-fastapi` [customizes the `FastAPI.openapi` method](https://github.com/stac-utils/stac-fastapi/blob/5e1ea82bd6be86e71689c4194e0e96f5c656723a/stac_fastapi/api/stac_fastapi/api/app.py#L331-L341) and does not pass the `servers` argument through, this section of the docs does not get autogenerated as it should. 

The signature of `StacAPI.customize_openapi` should probably be updated to pass through all of the same arguments as the default [`FastAPI.openapi` method](https://github.com/tiangolo/fastapi/blob/master/fastapi/applications.py#L197-L211).

```python
def openapi(self) -> Dict[str, Any]:
    if not self.app.openapi_schema:
        self.openapi_schema = get_openapi(
            title=self.title,
            version=self.api_version,
            openapi_version=self.app.openapi_version,
            description=self.description,
            terms_of_service=self.app.terms_of_service,
            contact=self.app.contact,
            license_info=self.app.license_info,
            routes=self.app.routes,
            tags=self.app.openapi_tags,
            servers=self.app.servers,
        )
    return self.app.openapi_schema
```

vincentsarago:
This was solved in 3.0

PR #499, opened 2022-11-25
OpenAPI construction

c-wygoda:
**Related Issue(s):** 

- #

**Description:**

- Use standard FastAPI OpenAPI builder and inject metadata through env settings
- Remove `stac_api.api.app:StacAPi.customize_openapi` method as it unnecessarily wraps the `FastApi.openapi` call.

**PR Checklist:**


geospatial-jeff:
@c-wygoda Whats the reason for this change?

c-wygoda:
We want to be able to specify more of the OpenApi metadata fields and in the process I found the customize_openapi method to be a rather redundant wrapper around the OpenApi functionality, if the FastAPI params are used directly in the attribute definition of the `app`.

c-wygoda:
This would also tackle #454 if I'm not mistaken.

geospatial-jeff:
> I found the customize_openapi method to be a rather redundant wrapper around the OpenApi functionality,

I agree, this code is leftover from years ago when customization of OpenAPI was less flexible.  Now `customize_openapi` isn't hurting more than it helps (ref #454).

> We want to be able to specify more of the OpenApi metadata fields

There is precedence to define openapi related things in `ApiSettings` so I think this is a good approach 👍 

Note to self that at some point in the future we should deprecate `StacApi.title`, `api_Version`, and `description` (obviously a breaking change).

geospatial-jeff (approved):

c-wygoda:
Thanks. I'm happy to help as I'm responsible for an instance, so hopefully can contribute more. Direct feedback welcome as I slowly learn the ropes of this project...

geospatial-jeff:
@c-wygoda The changes LGTM but I do want to play with this a little more to understand the difference between setting `servers` to `self.app.servers` compared to pulling from the environment within the context of #454 and #489

c-wygoda:
How can I help to move forward? More test coverage?

gadomski:
@geospatial-jeff ping

vincentsarago:
@jonhealy1 I think we should use this instead of #639 (we would need to update the landing page)

vincentsarago:
I'm not 💯 what we should do. 

This kinda breaking change so definitely something we should keep for 3.0, but at the same time #639 does almost the same thing without breaking anything. 

Both PR fixes stuff but do not represent the best solution to the issue:
- 499: fixes the OpenAPI but does not forward title, description,... to the landing page 
- 639: add info to both the OpenAPI and landing page but doesn't use the Config class and add some code duplication 

IMO we should wait after 2.5 to be released and see if we can refactor a bit to accommodate both changes 

cc @jonhealy1

m-mohr:
This PR should add some documentation how to set the env variables...

vincentsarago:
in #639 we implemented part of the proposed solution, Let's wait after 2.5 release to `finish` this PR

vincentsarago:
implemented most of this in https://github.com/stac-utils/stac-fastapi/pull/793

IMO adding more configuration is kinda `niche` and will be better handled by user themself (by creating a custom FastAPI application and passing it as argument to the StacApp class)

Issue #497, opened 2022-11-23
Inclusion of root-path prefix in paths available from service-desc when hosted behind a proxy


jfbourgon:
**Current and expected behavior**
As discussed in #489, when STAC-fastapi is hosted behind a proxy and started with a command like the following:
```
uvicorn --root-path "/stac" --proxy-headers stac_fastapi.sqlalchemy.app:app
```
All the links rendered by the application are the ones expected and they can be used to reach the various endpoints supported by the application. The only notable exception is inside the service-desc located under /stac/api, this endpoint is reachable but its content provides broken links since none of them include the expected root-path prefix (/stac in our specific case).

**Current**
```
{
	"openapi": "3.0.2",
	"info": {
		"title": "stac-fastapi",
		"version": "0.1"
	},
	"paths": {
		"/": {
			"get": {
				"summary": "Landing Page",
				"description": "Endpoint.",
				"operationId": "Landing_Page__get",
				"responses": {
					"200": {
						"description": "Successful Response",
						"content": {
							"application/json": {
								"schema": {}
							}
						}
					}
				}
			}
		},
		"/conformance": {
			"get": {
				"summary": "Conformance Classes",
				"description": "Endpoint.",
				"operationId": "Conformance_Classes_conformance_get",
				"responses": {
					"200": {
						"description": "Successful Response",
						"content": {
							"application/json": {
								"schema": {}
							}
						}
					}
				}
			}
		},
		"/collections/{collection_id}/items/{item_id}": {
			"get": {
				"summary": "Get Item",
				"description": "Endpoint.",
				"operationId": "Get_Item_collections__collection_id__items__item_id__get",
				"parameters": [
...
```

**Expected**
```
{
	"openapi": "3.0.2",
	"info": {
		"title": "stac-fastapi",
		"version": "0.1"
	},
	"paths": {
		"/stac/": {
			"get": {
				"summary": "Landing Page",
				"description": "Endpoint.",
				"operationId": "Landing_Page__get",
				"responses": {
					"200": {
						"description": "Successful Response",
						"content": {
							"application/json": {
								"schema": {}
							}
						}
					}
				}
			}
		},
		"/stac/conformance": {
			"get": {
				"summary": "Conformance Classes",
				"description": "Endpoint.",
				"operationId": "Conformance_Classes_conformance_get",
				"responses": {
					"200": {
						"description": "Successful Response",
						"content": {
							"application/json": {
								"schema": {}
							}
						}
					}
				}
[...]

PR #503, opened 2022-11-28
Add nginx service as second docker-compose stack

geospatial-jeff:
**Related Issue(s):** 

- #454
- #489
- #497

**Description:**
Adds a NGINX container to the docker-compose stack to help debug issues with root paths and link resolution when running the application behind a proxy; and to serve as a minimal reference example for how proxies should be configured.  The NGINX server runs on port `80` and proxies both backends behind the `/api/v1/pgstac` and `/api/v1/sqlalchemy` endpoints, root paths are set accordingly in each application.  I added the `UVICORN_ROOT_PATH` environment variable to expose uvicorn's root path argument without rewriting docker commands.

**PR Checklist:**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,16 +1,5 @@
 version: '3'
 services:
-  nginx:
-    image: nginx
-    ports:
-      - ${STAC_FASTAPI_NGINX_PORT:-80}:80
-    volumes:
-      - ./nginx.conf:/etc/nginx/nginx.conf
-    depends_on:
-      - app-pgstac
-      - app-sqlalchemy
-    command: ["nginx-debug", "-g", "daemon off;"]
-
   app-sqlalchemy:
     container_name: stac-fastapi-sqlalchemy
     image: stac-utils/stac-fastapi
@@ -30,7 +19,6 @@ services:
       - POSTGRES_HOST_WRITER=database
       - POSTGRES_PORT=5432
       - WEB_CONCURRENCY=10
-      - UVICORN_ROOT_PATH=/api/v1/sqlalchemy
     ports:
       - "8081:8081"
     volumes:
@@ -62,7 +50,6 @@ services:
       - DB_MIN_CONN_SIZE=1
       - DB_MAX_CONN_SIZE=1
       - USE_API_HYDRATE=${USE_API_HYDRATE:-false}
-      - UVICORN_ROOT_PATH=/api/v1/pgstac
     ports:
       - "8082:8082"
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,16 @@
 version: '3'
 services:
+  nginx:
+    image: nginx
+    ports:
+      - 80:80
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf
+    depends_on:
+      - app-pgstac
+      - app-sqlalchemy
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -1,5 +1,16 @@
 version: '3'
 services:
+  nginx:
+    image: nginx
+    ports:
+      - 80:80
```
c-wygoda, line 6: Once in a while, port 80 on the docker host may be occupied, allowing an override by env var might be helpful?
```suggestion
      - ${STAC_FASTAPI_NGINX_PORT:-80}:80
```

vincentsarago (approved):

docker-compose.yml
vincentsarago: does it mean you also need to change https://github.com/stac-utils/stac-fastapi/blob/17305101bd656e2d730affc983a815db31c01541/nginx.conf#L5 ?

c-wygoda: No, the first number is the port to bind to on the docker host, the second one the port in the container, so that would still be Nginx's port 80.

One potential gotcha is when full URLs are built in the actual service, it needs a way to figure out the external port.

docker-compose.yml
```diff
@@ -51,6 +63,7 @@ services:
       - DB_MIN_CONN_SIZE=1
       - DB_MAX_CONN_SIZE=1
       - USE_API_HYDRATE=${USE_API_HYDRATE:-false}
+      - UVICORN_ROOT_PATH=/api/v1/pgstac
```
vincentsarago, line 66: People might find a bit complicated to have to use the `nginx` service

you could `in theory` do something like 👇 
```yml
  nginx-sqlalchemy:
    extends:
      service: app-sqlalchemy
    ports:
    environment:
      - UVICORN_ROOT_PATH=/api/v1/sqlalchemy

  nginx:
    image: nginx
    ports:
      - 80:80
    volumes:
      - ./nginx.conf:/etc/nginx/nginx.conf
    depends_on:
      - [app](nginx-sqlalchemy:)
    command: ["nginx-debug", "-g", "daemon off;"]
```

vincentsarago: oh, I don't think you can `extend` services that have already `depends`.

> Keep in mind that volumes_from and depends_on are never shared between services using extends. These exceptions exist to avoid implicit dependencies; you always define volumes_from locally. This ensures dependencies between services are clearly visible when reading the current file. Defining these locally also ensures that changes to the referenced file don’t break anything.

you may need to re-add this

geospatial-jeff: Good point, this does break the links on the existing apps when going around nginx.  I think maybe we should just add a second compose stack - `docker-compose-nginx-proxy.yml`.

gadomski: @vincentsarago I picked this up and implemented as an optional docker-compose override file. I re-requested your review -- I tested locally but could use a second set of eyes.

vincentsarago: I think @geospatial-jeff would need to 👍 this too

gadomski: @geospatial-jeff ping -- if you don't mind taking a look, we can clear this PR out before 2.4.4. Thanks!

moradology (approved):
+1 - Tested new compose stack, rebased out conflicts, and updated changelog to reflect this feature

Issue #363, opened 2022-02-27
Declare non-200 responses in OpenAPI docs


geospatial-jeff:
https://fastapi.tiangolo.com/advanced/additional-responses/#additional-response-with-model

PR #523, opened 2023-02-06
Rework update_openapi function.

eseglem:
**Related Issue(s):** 

- I was thinking #363 might be an good place to start contributing but that ended up leading me to here while I was familiarizing myself with the code. 

**Description:**
The current implementation of `update_openapi` duplicates code from [fastapi/applications.py#L216](https://github.com/tiangolo/fastapi/blob/18d087f9c66b06c8baa4164bdf647af362b4a4ee/fastapi/applications.py#L216) without context and could end up out of sync with the source. Rather than duplicating code this implementation just wraps it and modifies the response's content-type header. 

While there are tests for the content-type header in other portions of the repo, there was not one in `stac_fastapi/api/tests`. To ease debugging I added a test to have one closer to the code that implements it.

I also removed the code that was marked with a `TODO` to remove or fix it. It was not being utilized and it seems like [stac_fastapi/api/app.py#L311](https://github.com/stac-utils/stac-fastapi/blob/master/stac_fastapi/api/stac_fastapi/api/app.py#L311) would be the place to do any changes in the future. While I did just notice #499 would remove that function for now, I still believe that would be the appropriate location if modifications to the schema are necessary in the future. 

**PR Checklist:**

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/openapi.py
+++ b/stac_fastapi/api/stac_fastapi/api/openapi.py
@@ -11,6 +11,20 @@ from stac_fastapi.api.config import ApiExtensions
 from stac_fastapi.types.config import ApiSettings
 
 
+class VndOaiResponse(JSONResponse):
+    """JSON with custom, vendor content-type."""
+
+    media_type = "application/vnd.oai.openapi+json;version=3.0"
+
+    def __init__(self, *args, **kwargs):
+        """Init function with deprecation warning."""
+        warnings.warn(
+            "VndOaiResponse is deprecated and will be removed in v3.0",
+            DeprecationWarning,
+        )
+        super().__init__(*args, **kwargs)
+
+
 def update_openapi(app: FastAPI) -> FastAPI:
     """Update OpenAPI response content-type.
 
@@ -42,20 +56,6 @@ def update_openapi(app: FastAPI) -> FastAPI:
 
     # return the patched app
     return app
-
-
-class VndOaiResponse(JSONResponse):
-    """JSON with custom, vendor content-type."""
-
-    media_type = "application/vnd.oai.openapi+json;version=3.0"
-
-    def __init__(self, *args, **kwargs):
-        """Init function with deprecation warning."""
-        warnings.warn(
-            "VndOaiResponse is deprecated and will be removed in v3.0",
-            DeprecationWarning,
-        )
-        super().__init__(*args, **kwargs)
 
 
 def config_openapi(app: FastAPI, settings: ApiSettings):


stac_fastapi/api/stac_fastapi/api/openapi.py
```diff
@@ -1,68 +1,36 @@
...
 from fastapi import FastAPI
-from fastapi.openapi.utils import get_openapi
 from starlette.requests import Request
-from starlette.responses import JSONResponse
-
-from stac_fastapi.api.config import ApiExtensions
-from stac_fastapi.types.config import ApiSettings
-
-
-class VndOaiResponse(JSONResponse):
```
gadomski, line 11: Should be deprecated instead of removed, so we can include this fix in a MINOR or PATCH release.

stac_fastapi/api/stac_fastapi/api/openapi.py
```diff
@@ -1,68 +1,36 @@
...
+        route for route in app.router.routes if route.path == app.openapi_url
     )
-    # Add the updated openapi route
-    app.add_route(app.openapi_url, openapi, include_in_schema=False)
-    return app
 
-
-# TODO: Remove or fix, this is currently unused
-# and calls a missing method on ApiSettings
-def config_openapi(app: FastAPI, settings: ApiSettings):
```
gadomski, line 45: Should be deprecated instead of removed.

stac_fastapi/api/stac_fastapi/api/openapi.py
```diff
@@ -1,68 +1,36 @@
...
+        # Update the content type header in place
+        response.headers[
+            "content-type"
+        ] = "application/vnd.oai.openapi+json;version=3.0"
+        # Return the updated response
+        return response
+
+    # When a route is accessed the `handle` function will call `self.app`. So we can
+    # leave the `endpoint` the original function for use in the patched function and
+    # just update the `app` to use the patched function.
```
gadomski, line 32: Is this a documented use-case? If so, can you drop a link in a reply just so I can read up and make myself smart? Thanks!

gadomski (changes requested):
Looks good -- I left one inline comment requesting more documentation so I can feel better about my knowledge of how the patching is working.

I requested changes b/c I think it would be more appropriate to deprecate, rather than than remove, the unused functions and classes. While we don't have a published deprecation strategy for **stac-fastapi** (note to self: TODO), there's enough implementations of this repo in the wild that I'm cautious about _any_ breaking changes, however small. IMO deprecating both `VndOaiRespone` and `config_openapi`, with a plan to remove in v3.0, would be appropriate.

Thanks for the contribution!

stac_fastapi/api/stac_fastapi/api/openapi.py
eseglem: Unfortunately this is not something I found any documentation on. I had to work my way through fastapi / starlette to figure out the process. I was attempting to document my thought process so someone else can follow what is happening in case modifications are needed down the line. I may have summarized too succinctly. 

My first attempt was to just modify the `openapi_route.endpoint` which seemed liked the obvious place to start. Unfortunately, that was unsuccessful and required much more digging. You can even set `endpoint = None` and nothing changes. 

What I figured out:
- The `openapi_route` is a `starlette.Route`
- When it is instantiated it does some checks to see if `endpoint` is a function / method and then wraps it with `request_response` and sets that as the `app`
- When the route is called it will call the `handle` function.
- When `handle` is called it only uses `self.app` so the `endpoint` function has no bearing on the output. 

So I just called the existing `endpoint` from the patched function. Wrapped it with `request_response` and set that as the `app`. Thinking about it more, it might be slightly better to do `old_endpoint = openapi_route.endpoint` and call that. Then actually update the endpoint on the route to be the patched function. The `__eq__` function does use it to compare routes.

eseglem:
@gadomski Added the functions back and added some deprecation warnings to them. It seems unlikely to be necessary, but being cautious is reasonable. 

It doesn't prevent private repos from using them, but there are no repos outside of `stac-utils` that have any mention of `VndOaiResponse`. And the only other use of `config_openapi` is an earlier fork: https://github.com/c-core-labs/stac-api which contains the function itself. The `config_openapi` also calls a function which no longer exists, and it probably cannot be used without an error. 

I also made some minor tweaks to the patch to hopefully make it more clear. There are potentially other options, like a hybrid of the old implementation of this one. Still create a new route, but with this patched endpoint function. I am certainly biased, but I prefer this way because it requires the least changes and code so there are less things to go wrong.

geospatial-jeff (approved):
LGTM

gadomski (approved):

PR #525, opened 2023-02-07
Add backend-specific dockerfiles, publish images

### LINKED ISSUE #379: Dockerfile improvements, and and deployable version (comments and questions)

### LINKED ISSUE #485: Put docker image on GHCR
Currently there is an image available for `pgstac` on GHCR, i.e. `ghcr.io/stac-utils/pgstac`, but there does not appear to be an image for `stac-fastapi`, which would be helpful.

Do you all of plans of doing that?

I've created a lightweight pytest fixture set that runs stac-fastapi, so that I can test `pystac-client` calls against it, and this would simplify things.

gadomski:
**Related Issue(s):** 

- Closes #379 
- Closes #485

**Description:**

WIP, testing CI.

**PR Checklist:**

- [ ] Code is formatted and linted (run `pre-commit run --all-files`)


moradology (approved):
Looks good. I don't think I have the permissions necessary to enable public container publishing. Reminder that this should be sorted out prior to merge

gadomski:
Looks like it worked (from https://github.com/stac-utils/stac-fastapi/pkgs/container/stac-fastapi):

image

PR #524, opened 2023-02-07
Remove links from default include list

gadomski:
**Related Issue(s):** 

- Closes #395 
- Closes alternate solution #527 

This is one of two PRs that would fix https://github.com/stac-utils/stac-fastapi/issues/395; the other is https://github.com/stac-utils/stac-fastapi/pull/527. We should merge one and close the other.

**Description:**

The [fields extension](https://github.com/stac-api-extensions/fields) is used to include/exclude fields on items returned by `/search` (and `/collections/{cid}/items`). If a field is not present in `include` or `exclude`, the extension does not specify whether that field should be present on the response items (though it does [provide guidance](https://github.com/stac-api-extensions/fields#includeexclude-semantics)). Currently, `links` is included by default, even if it is not present in the `include` list. This PR changes that behavior to **exclude `links` by default** if there is an `include` list.

**sqlalchemy** uses `FieldsExtension.default_includes`, so its fix was pretty easy
-- I did have to add an explicit include+exclude check to the search function to ensure the fields extension was applied _only if_ either `include` or `exclude` was set. **pgstac** was a little trickier, because it doesn't use the `default_includes` list.

I did modify one existing test -- `test_app_fields_extension` in the **sqlalchemy** tests was assuming fields would be stripped just by providing a `collections` parameter to the query, which I think is incorrect -- I added an explicit `include` field to the test.

**PR Checklist:**

- [x] Code is formatted and linted (run `pre-commit run --all-files`)
- [x] Tests pass (run `make test`)
- [x] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/stac-fastapi/blob/master/CHANGES.md).

PR #527, opened 2023-02-13
Align fields extension behavior with extension suggestions

### LINKED ISSUE #395: PgSTAC: Fields extension includes links even if not listed in `includes`
This test currently fails:

```python
@pytest.mark.asyncio
async def test_field_extension_includes_only_id_and_collection(
    app_client, load_test_data, load_test_collection
):
    """Test POST search including id and collection"""
    test_item = load_test_data("test_item.json")
    resp = await app_client.post(
        f"/collections/{test_item['collection']}/items", json=test_item
    )
    assert resp.status_code == 200

    body = {"fields": {"include": ["id", "collection"]}}

    resp = await app_client.post("/search", json=body)
    resp_json = resp.json()
    assert set(resp_json["features"][0].keys()) == {"id", "collection"}
```
with

```
E       AssertionError: assert {'links', 'id', 'collection'} == {'id', 'collection'}
E         Extra items in the left set:
E         'links'
E         Full diff:
E         - {'id', 'collection'}
E         + {'links', 'id', 'collection'}
E         ?  +++++++++
```

Links shouldn't be included in the response if they aren't in the "include" list.

gadomski:
**Related Issue(s):** 

- Closes alternate solution #524
- Closes #395

This is one of two PRs that would fix #395; the other is #524. We should merge one and close the other (should happen automatically thanks to closing keywords).

**Description:**

This PR updates both backends to (mostly) follow the recommended field extension semantics, as laid out in https://github.com/stac-api-extensions/fields#includeexclude-semantics. In particular:

- Any `include` fields are merged with the default include set
- An empty `fields` attribute will return the default include set
- If a field is both included and excluded, it is included in the response (this is a change from previous behavior)
- We include `stac_version` in the default includes to ensure valid STAC items
  - This is incorrect in the recommendations, and will be updated by https://github.com/stac-api-extensions/fields/pull/4)
- **In a deviation from the recommended semantics**, we keep `collection` and `stac_extensions`, as (in this author's opinion) these are valuable and cheap

The recommendations are encapsulated in a `into_recommended` method, where the set operations are performed: https://github.com/stac-utils/stac-fastapi/compare/issues/395-align-include-with-suggestions?expand=1#diff-b9d01a5a656ba2bdee8dd6eb5462bf1b1602b01bceaafb431cff8c38969815d5R61-R76.

I've also updated the fields extension default to be `None`, which allows for detection of an empty fields attribute (e.g. `?fields=`). This was required to enable item 1 in https://github.com/stac-api-extensions/fields#includeexclude-semantics.

**PR Checklist:**

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -155,7 +155,7 @@ jobs:
           cache: pip
           cache-dependency-path: stac_fastapi/pgstac/setup.cfg
       - name: Install stac-fastapi and stac-api-validator
-        run: pip install ./stac_fastapi/api ./stac_fastapi/types ./stac_fastapi/${{ matrix.backend }}[server] stac-api-validator==0.4.1
+        run: pip install ./stac_fastapi/api ./stac_fastapi/types ./stac_fastapi/extensions ./stac_fastapi/${{ matrix.backend }}[server] stac-api-validator==0.4.1
       - name: Run migration
         if: ${{ matrix.backend == 'sqlalchemy' }}
         run: cd stac_fastapi/sqlalchemy && alembic upgrade head
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 * Updated CI to test against [pgstac v0.6.12](https://github.com/stac-utils/pgstac/releases/tag/v0.6.12) ([#511](https://github.com/stac-utils/stac-fastapi/pull/511))
 * Reworked `update_openapi` and added a test for it ([#523](https://github.com/stac-utils/stac-fastapi/pull/523))
 * Limit values above 10,000 are now replaced with 10,000 instead of returning a 400 error ([#526](https://github.com/stac-utils/stac-fastapi/pull/526))
+* Default field include and exclude behavior ([#527](https://github.com/stac-utils/stac-fastapi/pull/527))
 
 ### Removed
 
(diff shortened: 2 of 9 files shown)


mmcfarland:
I understand the rationale for making it easy to always return a valid STAC Item despite what you put in `include`, and it makes sense to adhere to spec suggestions so that users get similar behavior between implementations. However, I have workflows where all I really care about is the ID (or maybe ID + datetime), and I'll just include that attribute in `include` in order to get the smallest possible payload (which can save a considerable amount of size/time when there are a lot of assets and a large result count). 

In that case, I understand I shouldn't try to parse the results as Items. I also think it's a bit cumbersome to have to copy over fields I would now see in my results to `exclude`, and many users who don't read the spec would be unlikely to assume they could even do that. That said, my use case is an optimization and having default valid STAC items is less likely to break anything. I'm moderately in favor of keeping the intuitive behavior (sans links) from pgstac backend.

philvarner:
Additionally, the Field Extension explicitly says that it makes returning invalid Items legal, so it's a *desire* to have valid items returned rather than a *requirement*.

gadomski:
Converting to draft until https://github.com/stac-api-extensions/fields/pull/5 lands. Once it does, I think we'll use #524 as the base solution, but there could be ideas from this PR that would be useful to port over.

gadomski:
Closing as blocked by https://github.com/stac-api-extensions/fields/pull/5.

Issue #550, opened 2023-03-29
#FeatureRequest add pgadmin to docker-compose


romeokienzler:
Add pgadmin to docker-compose

romeokienzler:
I'll create a PR for this

PR #551, opened 2023-03-29
add pgadmin to docker compose

romeokienzler:
**Related Issue(s):**

- #550 

**Description:**
Add pgadmin to docker-compose (UX/DX improvement)

**PR Checklist:**


docker/servers.json
gadomski, line 1: What's this file for? Is this required for pgadmin support?

gadomski (commented):
I could be convinced otherwise, but my initial instinct is that this is out-of-scope for this repository. Users can always set up their own **pgadmin** instance against their **pgstac**, and I'm not sure we need to provide that for them.

If this is useful information, could it be provided via documentation instead of another infrastructure file? I'm just trying to reduce the number of things we need to keep up-to-date and tested in this repo.

If we do want to keep this new docker-compose file, we should probably add a CI step to spin it up and run a smoke test, just to make sure it's still valid.

docker/servers.json
romeokienzler: yes - this is the config file pgadmin reads containing the DB connection details

romeokienzler:
> Users can always set up their own pgadmin instance against their pgstac

Not trying to convince anybody here, I just needed it and wanted to share my work with others. If you want to contribute to this project (which I'm planning to) it is useful to have pgadmin handy - and running it on docker alongside with the rest involves some complex networking which I've taken care of.

So I let you decide - if you wanna have it as a 3rd option, I'm happy to include it in automated testing, if not we can close this PR

Either way is fine for me

docker/servers.json
```diff
@@ -0,0 +1,22 @@
+{
+    "Servers": {
+        "1": {
+            "Name": "stac",
+            "Group": "Servers",
+            "Host": "pgstac,
```
gadomski, line 6:
```suggestion
            "Host": "pgstac",
```

gadomski (changes requested):

rajadain:
My 2 cents: I agree with @gadomski, don't think the inclusion of pgAdmin is a natural fit here. I personally use TablePlus and Postico, I'm sure there's other alternatives out there. Because pgAdmin isn't fundamental to the purpose of this library, don't think it should be included (single purpose principle).

The guidance provided in this PR on how to connect such a Postgres client to the database is useful though. Perhaps this could go in a FAQ / Guides section of the documentation?

gadomski:
@romeokienzler so I think the best course of action would be to add these instructions to the documentation, rather than as infrastructure code. Would you like to do that in this PR, or should I close this and you can open a docs PR later if you'd like?

romeokienzler:
Hi @gadomski let me use this PR, will ping when ready...thanks a lot!

romeokienzler:
@gadomski @rajadain please have a look

README.md
```diff
@@ -165,3 +166,22 @@ make docker-shell-pgstac # or docker-shell-sqlalchemy
...
+      - .:/var/lib/pgadmin/storage/username_email.com/
+    environment:
+      - PGADMIN_DEFAULT_EMAIL=username@email.com
+      - PGADMIN_DEFAULT_PASSWORD=password
+    ports:
+      - "8081:80"
+    command: postgres -N 500
+    links:
+      - "pgstac"
+```
```
gadomski, line 187: Thanks! I think the text looks good ... but feels a little specific for the `README.md`. Can you add it to the `tips-and-tricks` page in the documentation? https://github.com/stac-utils/stac-fastapi/blob/main/docs/tips-and-tricks.md

docker/servers.json
gadomski, line 1: File shouldn't be removed since we're not providing any infrastructure, just documentation.

gadomski (changes requested):

jonhealy1:
@gadomski @vincentsarago @romeokienzler This is really valuable. I would use this but I think it should be moved to the stac-fastapi-pgstac repository now as this repo is database-agnostic.

vincentsarago:
@jonhealy1 yeah it makes sense

PR #556, opened 2023-04-04
Remove ceil on fastapi version

gadomski:
**Description:**

It was added for some reason, but the reason appears to be lost in time (e.g. cb76e2e25e2fe7c285299bbc690133815432fd9f). I don't think this needs a changelog entry.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/types/setup.py
+++ b/stac_fastapi/types/setup.py
@@ -6,7 +6,7 @@ with open("README.md") as f:
     desc = f.read()
 
 install_requires = [
-    "fastapi>=0.73.0,<0.87",
+    "fastapi>=0.73.0",
     "attrs",
     "pydantic[dotenv]",
     "stac_pydantic==2.0.*",
--- a/stac_fastapi/api/tests/test_api.py
+++ b/stac_fastapi/api/tests/test_api.py
@@ -34,12 +34,12 @@ class TestRouteDependencies:
                 ), "Unauthenticated requests should be rejected"
                 assert response.json() == {"detail": "Not authenticated"}
 
-                make_request = getattr(client, route["method"].lower())
                 path = route["path"].format(
                     collectionId="test_collection", itemId="test_item"
                 )
-                response = make_request(
-                    path,
+                response = client.request(
+                    method=route["method"].lower(),
+                    url=path,
                     auth=("bob", "dobbs"),
                     data='{"dummy": "payload"}',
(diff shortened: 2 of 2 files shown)


gadomski:
Oh yeah, that's why: https://github.com/stac-utils/stac-fastapi/actions/runs/4612394172/jobs/8153239900?pr=556#step:13:93. Looks like I just need to update the test, there's no functional breakages: https://github.com/tiangolo/fastapi/releases/tag/0.87.0.

stac_fastapi/types/setup.py
```diff
@@ -6,7 +6,7 @@
     desc = f.read()
 
 install_requires = [
-    "fastapi>=0.73.0,<0.87",
+    "fastapi>=0.73.0",
```
vincentsarago, line 9: FYI 0.95 breaks quite some stuff in TiTiler (also 0.94 broke stuff with the new starlette version)

gadomski: Can you point me at an example or two? Doesn't look like the test suite for **stac-fastapi** catches those problems as it currently stands ... it'd be nice to push the tests into here so we can fix as needed.

Unless it's just a **TiTiler** thing, and we should let implementations ceil **fastapi** where they need?

vincentsarago: 👍 It might be only a TiTiler thing for sure. Issues were:
- 0.94 updated starlette and now the `request.url_for` methods returns a URL object (it was a string before): https://github.com/developmentseed/titiler/pull/607
- 0.95 added the `Annotated` type support and refactored most of the `fastapi.params` which now breaks titiler because we used Optional Path parameter (https://github.com/tiangolo/fastapi/discussions/9278)

There has been also other issues reported in https://github.com/tiangolo/fastapi/discussions since the release.

Again, just an FYI so you don't spend too long if the tests were failing ... but they seem to pass 😄

gadomski: Awesome, thanks. I'm inclined to 🙈 on those problems for this repo, and leave it to implementers to bound their **fastapi** versions as they see fit (folks _should_ be pinning to a specific version of all the deps for deployed instances, anyways, right?!).

vincentsarago: > folks should be pinning to a specific version of all the deps for deployed instances, anyways, right?

🙈

vincentsarago (approved):

PR #574, opened 2023-05-11
Update request body types for Transaction Extension spec and client

mmcfarland:
**Description:**

A POST against an items endpoint should accept an Item or ItemCollection

https://github.com/radiantearth/stac-api-spec/blob/master/ogcapi-features/extensions/transaction/README.md#methods

Added literal values to the stac_types `type` attribute in order to allow fastapi a discriminator value for parsing the request body to the correct TypedDict. Without this change, ItemCollections were parsed as Items and lost the `features` attribute.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
@@ -18,7 +18,9 @@ from stac_fastapi.types.extension import ApiExtension
 class PostItem(CollectionUri):
     """Create Item."""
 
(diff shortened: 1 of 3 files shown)


gadomski (approved):
@mmcfarland LGTM. I added a pretty basic smoke test, I don't think it actually exercises much (maybe some of the typing changes) but it at least demonstrates what we're trying to do? Do you mind taking a look at the test and seeing if it makes sense and if you think it has any value?

mmcfarland:
Thanks @gadomski I think it definitely adds value. I ran the test against main and got the expected failures. Thanks for the assist!

PR #614, opened 2023-10-23
Allow specifying insert method for bulk transactions

edkeeble:
**Description:**
We have started adopting the bulk transactions extension for loading data into our pgstac backend, but are blocked by the fact that we occasionally need to re-ingest an updated item. The `bulk_item_insert` route does not provide a means of specifying alternate data loading methods (e.g. upsert, delsert, etc) and [stac-fastapi-pgstac only supports inserts via the `create_items` function](https://github.com/stac-utils/stac-fastapi-pgstac/blob/0159cdb83304a968bb8d3f2f4ad4043bdbfe2619/stac_fastapi/pgstac/transactions.py#L185). In the event of a conflict, it returns a ConflictError with no details regarding which of the bulk items caused the conflict, which makes it difficult to correct the issue.

In this PR, I propose adding an optional parameter to the request object for the bulk_item_insert route. This `method` parameter will allow clients to specify the use of `upsert` rather than the default `insert` behavior. I have created a [separate PR on stac-fastapi-pgstac](https://github.com/stac-utils/stac-fastapi-pgstac/pull/64) to support conditionally using the `upsert_items` function.

This change should not impact any stac-fastapi backends which do not support alternate loading methods, since they will ignore the optional param.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
@@ -1,5 +1,6 @@
 """Bulk transactions extension."""
 import abc
+from enum import Enum
 from typing import Any, Dict, List, Optional, Union
 
 import attr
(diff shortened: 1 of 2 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/third_party/bulk_transactions.py
```diff
@@ -11,10 +12,18 @@
 from stac_fastapi.types.extension import ApiExtension
 
 
+class BulkTransactionMethod(str, Enum):
+    """Bulk Transaction Methods."""
+
+    INSERT = "insert"
```
bitner, line 18: Should we include the ignore method as well?

bitner (approved):

gadomski (approved):

Issue #593, opened 2023-09-06
Update to pydantic 2.0


gadomski:
I hear it's faster 😉

keul:
Is this related to #588 ?

gadomski:
Yes, #588 would probably be a blocker for this issue.

eseglem:
As an FYI, I finally got around to updating [pycql2](https://github.com/eseglem/pycql2/tree/develop) to work with v2. If it would be useful for this update. Its in the develop branch right now and is passing all my tests + some hypothesis fuzzing. Had to do a few odd things to get v2 working, but think its in a decent spot. And happy to address any concerns with it if there are any, as well as provide any help integrating.

thomas-maschler:
support for pydantic 2 was added in [stac-pydantic 3.0.0](https://github.com/stac-utils/stac-pydantic/releases/tag/3.0.0)

thomas-maschler:
@bmcandr please see #625

bmcandr:
:wave: @thomas-maschler. Thanks! I was actually asking about 588 with respect to your PR. The discussion in https://github.com/stac-utils/stac-pydantic/pull/127 _suggests_ that it's no longer a blocker on this issue, but I haven't seen confirmation. Just eager to get your PR merged 😄

vincentsarago:
done in #625

Issue #461, opened 2022-09-20
Response models are not being used to validate responses


drnextgis:
I've got confused by the fact that response models are not being used to validate responses. Isn't that its main purpose? As you can see [here](https://github.com/stac-utils/stac-fastapi/blob/b880bb3c8cbe3c282b6ac50788fe31bee575e536/stac_fastapi/api/stac_fastapi/api/app.py#L264) for example response model is [Collection](https://github.com/stac-utils/stac-pydantic/blob/86f7303efce7d16261d37c0d8ee333f6cc22e7ed/stac_pydantic/collection.py#L43). But if we modify collection's content in the database (pgstac) in order not to be align with the response model stac-fastapi simple ignores it and returns what we have in the database without any warning/error:
```
$ postgis=# UPDATE collections SET content = JSONB_SET(content, '{type}', '"MyCollection"');
```

```
$ curl -s -X 'GET' \
  'http://localhost:8082/collections/joplin' \
  -H 'accept: application/json' | jq '.type'
"MyCollection"
```

But according to the response model [type](https://github.com/stac-utils/stac-pydantic/blob/86f7303efce7d16261d37c0d8ee333f6cc22e7ed/stac_pydantic/collection.py#L55) should be `Collection`.

`ENABLE_RESPONSE_MODELS` set to `true`

gadomski:
I believe you're correct -- I've opened https://github.com/stac-utils/stac-fastapi/pull/519 to track the issue. Currently I just have a failing test to demonstrate the issue.

vincentsarago:
I confirmed this. 

The thing was that we are using `wrap_response` to make sure the endpoints return a starlette Response, but this issue is then FastAPI will not run pydantic validation/serialization if the endpoint respond a Response object 

https://github.com/stac-utils/stac-fastapi/blob/4fb10ec6ba758f28ddef749ea7bac6fbb5963f9b/stac_fastapi/api/stac_fastapi/api/routes.py#L18-L24


I've changed this in https://github.com/thomas-maschler/stac-fastapi/pull/2 for next stac-fastapi major version and will also change this for ~2.4 (maybe 2.5) https://github.com/stac-utils/stac-fastapi/pull/650#issuecomment-2042798205

vincentsarago:
I believe this is not fixed and tested

PR #625, opened 2024-02-10
update to pydantic 2

thomas-maschler:
**Related Issue(s):**

- #593 

**Description:**
This PR updates dependencies to Pydantic 2 and Stac-Pydantic 3
It removes the internal Stac, Search and Opertator Types in favor of Types from Stac Pydantic

**PR Checklist:**


lossyrob:
@thomas-maschler , moving the STAC types away from the TypeDict implementations in stac_fastapi.types to Pydantic types could have performance implications. Specifically I'm worried about the STAC types encoding Client output. These were originally moved away from Pydantic models as they tended to be slow for large payloads, e.g. returning an ItemCollection from search that contains 1000 large Sentinel 2 Items. Running Pydantic validation against known valid Items that come out of PgSTAC as part of response causes additional latency without a lot of value. I know the Pydantic 2 backend changes means validation will be faster, but are we introducing the potential for increased latency in the API routes because of the move away from TypedDicts in e.g. BaseCoreClient in this PR? If so, could the upgrade be done for but leave the TypeDict usage as is? Or have you seen that these changes don't introduce additional latency due to the magic of Rust 🪄?

thomas-maschler:
I can see your concerns regarding latency. I did not benchmark this myself. Others reported a [speed improvement of x5](https://thedataquarry.com/posts/why-pydantic-v2-matters/#results) between pydantic 1 and 2. I was assuming that this would be enough to address latency concerns.

The benefits I saw in keeping the pydandic models are peace of mind, due to basic checks and setting of default values you wouldn't have to store in the DB. I have a use case, where data are stored in a legacy system and I have to translate them to STAC at runtime. Being able to push part of this into Pydantic makes the job easier.

If latency is still the main concern I could see several routes

1. Keep TypedDicts
Be explicit in the documentation that responses are not type checked and it is the responsibility of the use to do so. 

2. Allow to skip validation for ItemCollections and Collections

```
from pydantic import SkipValidation
from stac_pydantic import api
from typing import Sequence

class ItemCollectionUnsafe(api.ItemCollection):
    features: Sequence[SkipValidation[api.Item]]
 
class CollectionsUnsafe(api.Collections):
    collections: Sequence[SkipValidation[api.Collection]]
```

With minimal code changes, users can use the `Unsafe` subclasses as their return model, telling pydantic to skip validation for items and collections when making bulk requests. 

3. Add some configuration options to let the user choose between current implementation options 1 or 2 above.

rhysrevans3:
Couple of notes from my pull request:
1. If typing `List` and `Dict` are being used [this part](https://github.com/stac-utils/stac-fastapi/pull/625/files#diff-f84457de825c3fa96fa65673a6f85442ac57bd10f9f939cb7411780d3b6fabb5R46-R47) will need to be changed as in it's current form it will miss off the `Optional` typing of nested `Lists` & `Dicts`.
2. The [types.links](https://github.com/stac-utils/stac-fastapi/blob/02f2702a64476ee63bca90a524ce9fc4d2cc91de/stac_fastapi/types/stac_fastapi/types/links.py) seems to be covered by stac-pydantic links and [link_factory](https://github.com/stac-utils/stac-pydantic/blob/main/stac_pydantic/api/utils/link_factory.py).
3. The stac-pydantic Item/Collection/ItemCollection types are not json serialisable so will need to be json encoded/converted to dictionaries before they're returned from the API.

thomas-maschler:
@rhysrevans3, regarding your first point, can you give me an example of when this fails? I cannot replicate it. I tried this, where the filter attribute is defined as `Optional[Dict]`

```python

@pytest.mark.parametrize(
    "filter,passes",
    [(None, True), ({"test": "test"}, True), ("test==test", False), ([], False)],
)
def test_create_post_request_model(filter, passes):
    extensions = [FilterExtension()]
    request_model = create_post_request_model(extensions, BaseSearchPostRequest)

    if not passes:
        with pytest.raises(ValidationError):
            model = request_model(filter=filter)
    else:
        model = request_model(
            collections=["test1", "test2"],
            ids=["test1", "test2"],
            bbox=[0, 0, 1, 1],
            datetime="2020-01-01T00:00:00Z",
            limit=10,
            filter=filter,
            **{"filter-crs": "epsg:4326", "filter-lang": "cql2-text"},
        )

        assert model.collections == ["test1", "test2"]
        assert model.filter_crs == "epsg:4326"
        assert model.filter == filter
```

lossyrob:
@thomas-maschler understood about the value of validation in the response classes for other use cases.

I think a solution where users could skip validation based on configuration would be great. In my previous testing (a year or so ago) it was the Pydantic validation that was causing slowdowns. I'd still want to validate that there's no other performance hits besides validation for using Pydantic models over TypeDicts before integrating these changes, but I have enough confidence that this would address the performance implications, and any unforeseen issues could be addressed in follow up work.

Would you be willing to add that SkipValidation configuration to this PR?

vincentsarago:
about performances, in [**TiPg**](https://github.com/developmentseed/tipg) we also went for the `TypedDict` way because pydantic validation was really slowing down the response (even with pydantic 2.0)

rhysrevans3:
@thomas-maschler I found it was an issue for nested List/Dicts types: for example the `sortby` field in the SortExtensionPostRequest is
```
sortby: Optional[List[PostSortModel]]
```
But when I check the `FieldInfo` shows `required=True`
```
>>> from stac_fastapi.extensions.core.sort.request import SortExtensionPostRequest
>>>SortExtensionPostRequest.model_fields['sortby']
FieldInfo(annotation=Union[List[SortExtension], NoneType], required=True)
```

thomas-maschler:
@rhysrevans3 this is b/c we didn't set a default value. Right now the model requires either `None` or a list of `PostSortModel`. If you set the default to `None` the FieldInfo will say `required=False`. However, `None` is filled in when `sortby` is not provided so it still works even without a default value.


Regarding your point 2.
The `type.Links` are no Pydantic Models and are not used anywhere in the package. I prefer to leave them in right now to not break too much. I am not sure if any of the backends currently rely on it.

Regarding your point 3.
All pydantic models are json serializable. either with `.model_dump_json()` (output as `str`) or `.model_dump(mode="json")` (output as `dict`). I added some test, directly returning the pydantic `Item`, `Collection` and `ItemCollection` and this works for me.

I am still working on the switch between `TypeDict` and `Pydantic` response models and will probably push my updates later today.

rhysrevans3:
@thomas-maschler ah that makes more sense thank you for the explanation. I guess this just need to be filled in the backend I'm using. 

Point 2 that sounds like a good idea. But don't the links in the [api.landing_page](https://github.com/stac-utils/stac-pydantic/blob/main/stac_pydantic/api/landing.py) used [here](https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/core.py#L260-L312) need to be stac-pydantic `Links`? I guess with the validation removed this won't be an issue.

Point 3 sorry yes I meant `json.dumps` or `orjson` won't serialise them which I think fastapi uses for returning JSON/GEOJSON content. So was more of a note that the other repos will need to serialise or convert to a `dict` before returning the api routes. I think this effect the [landing pages](https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/core.py#L354-L406). however, I might be missing something obvious.

thomas-maschler:
@lossyrob can you take a look at my recent changes, in particular [8118f10](https://github.com/stac-utils/stac-fastapi/pull/625/commits/8118f10be27f04d5ce2143b5fa9dbbd1b4fd92c3).

- In `types.config` I added a parameter `validate_response` that defaults to `False` (same as now).
- I added `types.stac` back in
- I added `types.response_model` which implements the switch
- In `api.core` I made sure we always use response_model classes and added some extra type-checking 
- I added additional tests to make sure this works correctly, in particular, `test_response_model` and `test_app`

I stumbled over some other bugs regarding field aliases in the filter extension that I didn't manage to fix. I will file an issue for that.
I also noticed that there are still some pydantic extension models that are either out of sync with stac-pydantic or not present in stac-pydantic. I will port them over and open another PR once they make their way through the system.

thomas-maschler:
@rhysrevans3 
regarding your point 3: please look at `test_app.py` in this PR. Serialization works just fine

rhysrevans3:
@thomas-maschler sorry about that I've been getting `TypeError: Object of type LandingPage is not JSON serializable` on the landing page when trying to update the [stac-fastapi-elasticsearch-opensearch](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch) to use this branch and assumed it was because of the switch to `api.LandingPage` type but this is my first time using pydantic models. I'll keep digging. Thanks for all your help!

thomas-maschler:
I haven't looked at the backends. We use a custom one.
Try importing `types.response_model.LandingPage` and see if it works now.

If you want to use the Pydantic models set the envvar `VALIDATE_RESPONSE=True`

vincentsarago:
Hi everyone, I'm going to take over the review. 

@thomas-maschler did you had anything to add or is this PR ready?

thomas-maschler:
Thanks, this is ready. The most controversial part is probably the response model

jonhealy1:
I added a comment about adding something in the readme explaining the option to choose the response model.

stac_fastapi/api/tests/test_app.py
```diff
@@ -0,0 +1,149 @@
...
+
+@pytest.mark.parametrize(
+    "validate, response_type",
+    [
+        ("True", BaseModel),
+        ("False", dict),
+    ],
+)
+def test_app(validate, response_type, collection_dict, item_dict, cleanup):
+    os.environ["VALIDATE_RESPONSE"] = str(validate)
```
vincentsarago, line 34: could we use monkeypatch here instead of writing the env variable 
https://docs.pytest.org/en/6.2.x/monkeypatch.html

stac_fastapi/api/tests/test_app.py
```diff
@@ -0,0 +1,149 @@
...
+            query: Optional[str] = None,
+            token: Optional[str] = None,
+            fields: Optional[List[str]] = None,
+            sortby: Optional[str] = None,
+            intersects: Optional[str] = None,
+            **kwargs,
+        ) -> response_model.ItemCollection:
+            # FIXME: hyphen alias for filter_crs and filter_lang are currently not working
+            # assert kwargs.get("filter_crs") == "EPSG:4326"
+            # assert kwargs.get("filter_lang") == "cql-test"
```
vincentsarago, line 62: I'm not sure to get this. Did you open an issue about it yet?

stac_fastapi/api/tests/test_app.py
```diff
@@ -0,0 +1,149 @@
+import importlib
```
vincentsarago, line 1: can we add some docstrings explaining what this `test_app` will test?

stac_fastapi/api/setup.py
```diff
@@ -7,8 +7,8 @@
 
 install_requires = [
     "attrs",
-    "pydantic[dotenv]<2",
-    "stac_pydantic==2.0.*",
+    "pydantic[dotenv]>=2",
+    "stac_pydantic>=3",
```
vincentsarago, line 11: we should maybe restrict stac_pydantic to `<4.0`

stac_fastapi/api/setup.py
```diff
@@ -7,8 +7,8 @@
 
 install_requires = [
     "attrs",
-    "pydantic[dotenv]<2",
-    "stac_pydantic==2.0.*",
+    "pydantic[dotenv]>=2",
```
vincentsarago, line 10: I think `dotenv` was removed in pydantic 2.0  

python-dotenv is now the dependency of pedantic-settings

stac_fastapi/api/stac_fastapi/api/models.py
```diff
@@ -44,40 +43,19 @@ def create_request_model(
...
-                    min_items=field_info.min_items,
-                    max_items=field_info.max_items,
-                    min_length=field_info.min_length,
-                    max_length=field_info.max_length,
-                    regex=field_info.regex,
-                    extra=field_info.extra,
-                )
-                fields[k] = (v.outer_type_, body)
+            for k, field_info in model.model_fields.items():
+                fields[k] = (field_info.annotation, field_info)
```
vincentsarago, line 47: can you elaborate why we changed this?

stac_fastapi/api/tests/test_models.py
```diff
@@ -0,0 +1,100 @@
...
+            {
+                "type": "Polygon",
+                "coordinates": [[[0, 0], [0, 1], [1, 1], [1, 0], [0, 0]]],
+            }
+        ),
+        datetime="2020-01-01T00:00:00Z",
+        limit=10,
+        filter="test==test",
+        # FIXME: hyphen aliases are not properly working
+        # **{"filter-crs": "epsg:4326", "filter-lang": "cql2-text"},
```
vincentsarago, line 30: do we know if this was working before pydantic 2.0 ?

vincentsarago: maybe this is linked to the changes at https://github.com/stac-utils/stac-fastapi/pull/625/files#diff-f84457de825c3fa96fa65673a6f85442ac57bd10f9f939cb7411780d3b6fabb5R47

stac_fastapi/api/setup.py
vincentsarago: maybe we should just remove the requirements here and rely on `stac-fastapi.types` requirements?

maybe we should also pin `stac-fastapi.types` version

stac_fastapi/extensions/setup.py
```diff
@@ -7,8 +7,8 @@
 
 install_requires = [
     "attrs",
-    "pydantic[dotenv]<2",
-    "stac_pydantic==2.0.*",
+    "pydantic[dotenv]>=2",
+    "stac_pydantic>=3",
```
vincentsarago, line 11: same for `api` 
- `dotenv` is not an option anymore
- limit `stac_pydantic` upper version
- pin `stac-fastapi.types` and `stac-fastapi.api` version

lossyrob:
> @lossyrob can you take a look at my recent changes, in particular [8118f10](https://github.com/stac-utils/stac-fastapi/pull/625/commits/8118f10be27f04d5ce2143b5fa9dbbd1b4fd92c3).
> 
> * In `types.config` I added a parameter `validate_response` that defaults to `False` (same as now).
> * I added `types.stac` back in
> * I added `types.response_model` which implements the switch
> * In `api.core` I made sure we always use response_model classes and added some extra type-checking
> * I added additional tests to make sure this works correctly, in particular, `test_response_model` and `test_app`
> 
> I stumbled over some other bugs regarding field aliases in the filter extension that I didn't manage to fix. I will file an issue for that. I also noticed that there are still some pydantic extension models that are either out of sync with stac-pydantic or not present in stac-pydantic. I will port them over and open another PR once they make their way through the system.

Thanks for this change, this for sure satisfies the concerns I brought up.

Looking forward to this making its way into a release soon!

stac_fastapi/api/tests/test_app.py
thomas-maschler: yep, https://github.com/stac-utils/stac-fastapi/issues/638

stac_fastapi/api/tests/test_models.py
thomas-maschler: hmm, let me test that

stac_fastapi/api/setup.py
thomas-maschler: Following [Henry Schreiner advice](https://iscinumpy.dev/post/bound-version-constraints/) I usually only add lower bounds.

stac_fastapi/api/tests/test_models.py
thomas-maschler: I tested locally and reverting back to using the fastapi.Body instead of FieldInfo doesn't change that.
This section only handles POST requests that use the Pydantic BaseModel.

The issue with aliases only appears with GET requests that use attrs classes with Pydantic Fields. I don't think that ever worked. I will do some more testing tomorrow.

stac_fastapi/api/setup.py
```diff
@@ -1,16 +1,20 @@
...
 
+from distutils.util import convert_path
+
 from setuptools import find_namespace_packages, setup
 
+main_ns = {}
+ver_path = convert_path("stac_fastapi/api/version.py")
+with open(ver_path) as ver_file:
+    exec(ver_file.read(), main_ns)
+
```
vincentsarago, line 11: Can we avoid doing this? 

I think it's fine to manually hard code the version here. 

We can then use something like `bump2version` to update all the files we want when doing a release

e.g. https://github.com/developmentseed/titiler/blob/main/.bumpversion.cfg

stac_fastapi/api/tests/test_app.py
```diff
@@ -30,8 +33,9 @@ def cleanup():
         ("False", dict),
     ],
 )
-def test_app(validate, response_type, collection_dict, item_dict, cleanup):
-    os.environ["VALIDATE_RESPONSE"] = str(validate)
+def test_app(validate, response_type, collection_dict, item_dict, cleanup, monkeypatch):
+    monkeypatch.setenv("VALIDATE_RESPONSE", validate)
```
vincentsarago, line 37: do we still need the `cleanup`?

stac_fastapi/types/stac_fastapi/types/response_model.py
```diff
@@ -0,0 +1,26 @@
...
+LandingPage = response_model.LandingPage
+Collection = response_model.Collection
+Collections = response_model.Collections
+Item = response_model.Item
+ItemCollection = response_model.ItemCollection
+try:
+    Conformance = response_model.Conformance
+except AttributeError:
+    # TODO: class name needs to be fixed in stac_pydantic
+    Conformance = response_model.ConformanceClasses
```
vincentsarago, line 26: ref: https://github.com/stac-utils/stac-pydantic/pull/136

stac_fastapi/types/stac_fastapi/types/search.py
```diff
@@ -1,79 +1,23 @@
...
-    # in = auto()
-
-    @DynamicClassAttribute
-    def operator(self) -> Callable[[Any, Any], bool]:
-        """Return python operator."""
-        return getattr(operator, self._value_)
+from pydantic import PositiveInt
+from pydantic.functional_validators import AfterValidator
+from stac_pydantic.api import Search
+from typing_extensions import Annotated
```
vincentsarago, line 12: There is a lot of changes in this file, can you explain?

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -306,10 +309,20 @@ def _landing_page(
                     "href": urljoin(base_url, "search"),
                     "method": "POST",
                 },
+                {
+                    "rel": Relations.service_desc.value,
+                    "type": MimeTypes.geojson,
```
vincentsarago, line 314: should this be 
```
"type": MimeTypes.geojson.value,
```

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -262,8 +265,8 @@ def _landing_page(
         base_url: str,
         conformance_classes: List[str],
         extension_schemas: List[str],
-    ) -> stac_types.LandingPage:
-        landing_page = stac_types.LandingPage(
+    ) -> Dict[str, Any]:
```
vincentsarago, line 268: Why is this method returning a Dict instead of `response_model.LandingPage` ?

vincentsarago:
Thanks a ton @thomas-maschler

This PR is `massive` and has many unrelated changes to pydantic update but it's too late to refactor so we'll just try to merge this. 

I've posted some comment directly in the files themself, but wanted to ask one `general` question. 

I see that we now have an option for the Model to be either TypeDict or Pydantic Model. This model is used  at two levels (if I understand well):

1. between the Backend and the Client (https://github.com/thomas-maschler/stac-fastapi/blob/pydantic2/stac_fastapi/types/stac_fastapi/types/stac.py#L490)

2. between the Client and the User (via FastAPI response) (https://github.com/thomas-maschler/stac-fastapi/blob/pydantic2/stac_fastapi/api/stac_fastapi/api/app.py#L238-L240) 

Question:
- I wonder if there is a real need to the Pydantic Model to be used at level 1, we could still get validation error when FastAPI would try to validate the response? 
- in https://github.com/thomas-maschler/stac-fastapi/blob/e24351ad2c182a790128d11423cc05870df7dd9f/stac_fastapi/api/stac_fastapi/api/app.py#L238-L240
if `self.settings.enable_response_models` is set to `False` we won't have OpenAPI schema hints for the response. In `TiPG` we overcome this by passing the model directly to the OpenAPI like https://github.com/developmentseed/tipg/blob/main/tipg/factory.py#L688 . Do you think we could do the same here (we can do it in another PR)

stac_fastapi/types/stac_fastapi/types/search.py
thomas-maschler: I removed classes that are now implemented in stac_pydantic

stac_fastapi/api/stac_fastapi/api/models.py
thomas-maschler: I am not sure why we used fastapi.Body in first place, but some API changes in pydantic 2 didn't allow to do a 1:1 mapping between `FieldInfo` and `Body`. However, you can now pass a `dict` of `FieldInfo` objects to `create_model` and it will populate the model for you. This is much cleaner than the previous solution.

stac_fastapi/api/tests/test_models.py
thomas-maschler: I did some more testing and I am confident that this bug already existed before my changes and I only discovered it. I added several tests and linked them back to the issue I created for the bug.

stac_fastapi/api/setup.py
thomas-maschler: ok, i will hard code it for now.
I created a follow up ticket
https://github.com/stac-utils/stac-fastapi/issues/648

stac_fastapi/types/stac_fastapi/types/core.py
thomas-maschler: This is an internal method and we later modify the dictionary further. Instead of adding toggles between the two response types throughout the code, it is easier to default to a `dict` here and do a final cast when returning the response. Returning a `stac.LandingPage` would also work if that adds more clarity

thomas-maschler:
Oh, I somehow missed out on option 2. That makes things a lot clearer now. 
Let's go with that approach only. It is less convoluted than what I did. 

I will revert my introduced changes and make sure FastAPI uses the correct models.
Let me also see how I can inject the model to OpenAPI

thomas-maschler:
ok, I got this reverted but I don't get my tests to pass. Am I doing something wrong?
Even if I set [`enable_response_models` to `True`](https://github.com/stac-utils/stac-fastapi/pull/625/commits/eea9c8227705a35425b7e2f864946a787f9ecaef#diff-654b52c9fb869f6f11556091a7f2814122d2c88ec65453d168a85c26516da5edR81) I get a `200` response with an invalid item.

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -383,10 +379,8 @@ def landing_page(self, **kwargs) -> response_model.LandingPage:
 
         # Add Collections links
         _collections = self.all_collections(request=kwargs["request"])
-        if isinstance(_collections, BaseModel):
-            collections = _collections.model_dump(mode="json")
-        else:
-            collections = _collections
+        collections = _collections
```
vincentsarago, line 382: I don't think we need to create a new variable here! 

```python
collections = self.all_collections(request=kwargs["request"])
for collection in collections["collections"]:
```

vincentsarago:
> ok, I got this reverted but I don't get my tests to pass. Am I doing something wrong?
Even if I set [enable_response_models to True](https://github.com/stac-utils/stac-fastapi/pull/625/commits/eea9c8227705a35425b7e2f864946a787f9ecaef#diff-654b52c9fb869f6f11556091a7f2814122d2c88ec65453d168a85c26516da5edR81) I get a 200 response with an invalid item.

🤯 this is so weird!

vincentsarago:
```python

import pytest

from pydantic import BaseModel
from fastapi import FastAPI
from fastapi.exceptions import ResponseValidationError
from fastapi.testclient import TestClient


class model(BaseModel):
    value: int


def test_model():
    app = FastAPI()

    # should return 200
    @app.get(
        "/yo",
        response_model=model,
        responses={
            200: {
                "content": {
                    "application/json": {},
                },
                "model": model,
            },
        },
    )
    def yo():
        return {"value": 1}

    # response model
    # so validation should fail
    @app.get(
        "/ye",
        response_model=model,
        responses={
            200: {
                "content": {
                    "application/json": {},
                },
                "model": model,
            },
        },
    )
    def ye():
        return {"value": "b"}

    # No response model
    # so No validation
    # should return 200
    @app.get(
        "/ya",
        responses={
            200: {
                "content": {
                    "application/json": {},
                },
                "model": model,
            },
        },
    )
    def ya():
        return {"value": "b"}

    # SAME With `add_api_route` method
    def yi():
        return {"value": "b"}

    app.router.add_api_route(
        name="yi",
        path="/yi",
        response_model=model,
        responses={
            200: {
                "content": {
                    "application/json": {},
                },
                "model": model,
            },
        },
        methods=["GET"],
        endpoint=yi,
    )

    with TestClient(app) as client:
        resp = client.get("/yo")
        assert resp.status_code == 200

        with pytest.raises(ResponseValidationError):
            client.get("/ye")

        resp = client.get("/ya")
        assert resp.status_code == 200

        with pytest.raises(ResponseValidationError):
            client.get("/yi")
```

I've tried to reproduce this with a simple APP but without success 😬 , ☝️ pass the tests

vincentsarago:
ahhhhhh it's because the endpoint response get's automatically wrapped in a `Response` so it passthrough the validation 😭 

https://github.com/tiangolo/fastapi/blob/7e161b3f9e7ae5b6a37ab8ea13b6dd2c1c11eff9/fastapi/routing.py#L281-L284

so 👇 won't raise an issue 
```
    @app.get(
        "/yu",
        response_model=model,
        responses={
            200: {
                "content": {
                    "application/json": {},
                },
                "model": model,
            },
        },
    )
    def yu():
        return JSONResponse({"value": "b"})
```

vincentsarago:
I'm pretty we can remove this 

https://github.com/thomas-maschler/stac-fastapi/blob/d0762eb20dfba437376c9a8879524cac0ee9ce7e/stac_fastapi/api/stac_fastapi/api/routes.py#L26

because FastAPI will automatically do it, I'll make a PR on top of your PR

stac_fastapi/extensions/stac_fastapi/extensions/core/context.py
```diff
@@ -1,4 +1,5 @@
 """Context extension."""
```
vincentsarago, line 1: The context extension is deprecated! I think we can 🔪 it!

vincentsarago: maybe in another PR

thomas-maschler:
Phew, that was a beast. It kind of grew out of hand. Thanks for the help.
One thing left is updating the documentation. I believe, that once I turn that stone, new things will pop up and I prefer to make this a separate PR.

vincentsarago:
thanks a lot @thomas-maschler. 

One thing that I want to do before we merge is make sure the performance is 👍, I'll try to set up some benchmark over the weekend

stac_fastapi/extensions/stac_fastapi/extensions/core/context.py
thomas-maschler: https://github.com/stac-utils/stac-fastapi/issues/649

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -437,15 +446,11 @@ def get_search(
         collections: Optional[List[str]] = None,
         ids: Optional[List[str]] = None,
         bbox: Optional[List[NumType]] = None,
+        intersects: Optional[str] = None,
         datetime: Optional[Union[str, datetime]] = None,
         limit: Optional[int] = 10,
-        query: Optional[str] = None,
-        token: Optional[str] = None,
-        fields: Optional[List[str]] = None,
-        sortby: Optional[str] = None,
```
jonhealy1, line 445: What are the reasons for removing these fields?

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -633,15 +640,11 @@ async def get_search(
...
```
jonhealy1, line 641: Same question here.

stac_fastapi/types/stac_fastapi/types/config.py
```diff
@@ -30,11 +31,7 @@ class ApiSettings(BaseSettings):
     openapi_url: str = "/api"
     docs_url: str = "/api.html"
 
-    class Config:
-        """Model config (https://pydantic-docs.helpmanual.io/usage/model_config/)."""
-
-        extra = "allow"
-        env_file = ".env"
+    model_config = SettingsConfigDict(env_file=".env", extra="allow")
```
jonhealy1, line 34: I think it would be nice to have a small section in the readme explaining how to set the response validation.

jonhealy1 (changes requested):
Everything looks really good to me. A couple questions and a request to add an explanation about the `enable_response_models` option to the readme. Also interested in any performance testing @vincentsarago is doing.

stac_fastapi/types/stac_fastapi/types/core.py
thomas-maschler: They are used for extensions only and should not be part of this abstract method. Users can implement them in their own subclass if they need to.

stac_fastapi/types/stac_fastapi/types/core.py
thomas-maschler: Same reason as above

stac_fastapi/types/stac_fastapi/types/core.py
jonhealy1: Makes sense. Thank you.

drnextgis:
Do you think [this](https://github.com/stac-utils/stac-fastapi/issues/461) issue is going to be fixed once this PR is merged?

vincentsarago:
@drnextgis Hey sorry for not seeing your ticket before, but yes it will be fixed. I stumbled on this to https://github.com/thomas-maschler/stac-fastapi/pull/2

I'll fix also for ~2.4 version of stac-fastapi

vincentsarago:
```
----------------------------------- benchmark 'Items': 7 tests -----------------------------------
Name (time in ms)           Min                Max               Mean             Median          
--------------------------------------------------------------------------------------------------
Items Limit: (1)         1.0452 (1.0)       2.1031 (1.0)       1.1329 (1.0)       1.1137 (1.0)    
Items Limit: (10)        1.7471 (1.67)      3.0766 (1.46)      1.8286 (1.61)      1.8087 (1.62)   
Items Limit: (50)        4.5001 (4.31)     64.0863 (30.47)     4.9263 (4.35)      4.6144 (4.14)   
Items Limit: (100)       8.0186 (7.67)     64.3674 (30.61)     8.6762 (7.66)      8.1762 (7.34)   
Items Limit: (200)      14.8327 (14.19)    17.0791 (8.12)     15.1425 (13.37)    15.0344 (13.50)  
Items Limit: (250)      18.1781 (17.39)    58.4482 (27.79)    19.1771 (16.93)    18.4300 (16.55)  
Items Limit: (1000)     69.1144 (66.13)    69.9298 (33.25)    69.6401 (61.47)    69.6901 (62.58)  
--------------------------------------------------------------------------------------------------

-------------------------------- benchmark 'Items With Model validation': 7 tests --------------------------------
Name (time in ms)                          Min                 Max               Mean             Median          
------------------------------------------------------------------------------------------------------------------
Items With Model validation (1)         1.1480 (1.0)        2.3425 (1.0)       1.2314 (1.0)       1.2174 (1.0)    
Items With Model validation (10)        1.7948 (1.56)       2.8828 (1.23)      1.9552 (1.59)      1.9323 (1.59)   
Items With Model validation (50)        4.1826 (3.64)      38.1875 (16.30)     4.7179 (3.83)      4.2844 (3.52)   
Items With Model validation (100)       7.2982 (6.36)      43.5653 (18.60)     8.6554 (7.03)      7.4836 (6.15)   
Items With Model validation (200)      12.8544 (11.20)     47.9423 (20.47)    17.3208 (14.07)    13.2296 (10.87)  
Items With Model validation (250)      16.0176 (13.95)     53.0070 (22.63)    21.2084 (17.22)    16.2596 (13.36)  
Items With Model validation (1000)     61.4995 (53.57)    106.9911 (45.67)    86.3199 (70.10)    99.1915 (81.48)  
------------------------------------------------------------------------------------------------------------------
```

Well, either the validation/serialization is not working or there is no difference 🤯

jonhealy1:
@vincentsarago I thought we had results [here](https://github.com/stac-utils/stac-fastapi/pull/650#issuecomment-2042918120) that seemed to make sense?

thomas-maschler:
I believe we now test for correct validation.

vincentsarago:
> I believe we now test for correct validation.

@thomas-maschler Yes

> @vincentsarago I thought we had results https://github.com/stac-utils/stac-fastapi/pull/650#issuecomment-2042918120 that seemed to make sense?

@jonhealy1 absolutely, so this mean that validation does not impact the performance 🤯. I would still keep this optional but it might be great to mention this somewhere

thomas-maschler:
Ok, I'll update the readme

stac_fastapi/types/stac_fastapi/types/config.py
thomas-maschler: I added some language to the Readme

jonhealy1: Looks great. Thank you

vincentsarago:
thanks @thomas-maschler I was going update this branch tomorrow morning 🙏 

let see if we can get this merged before the end of the week 😄

stac_fastapi/types/stac_fastapi/types/version.py
```diff
@@ -1,2 +1,3 @@
 """Library version."""
+
 __version__ = "2.5.4"
```
jonhealy1, line 3: 2.5.5

jonhealy1 (changes requested):
Nice work. The readme looks great - just a couple of spelling things

stac_fastapi/extensions/setup.py
```diff
@@ -1,16 +1,14 @@
...
     desc = f.read()
 
 install_requires = [
-    "attrs",
-    "pydantic[dotenv]<2",
-    "stac_pydantic==2.0.*",
-    "stac-fastapi.types",
-    "stac-fastapi.api",
+    "stac-fastapi.types==2.4.9",
+    "stac-fastapi.api==2.4.9",
```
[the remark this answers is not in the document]
vincentsarago: not really because we will do #648 ;-)

stac_fastapi/api/setup.py
```diff
@@ -57,4 +54,5 @@
     install_requires=install_requires,
     tests_require=extra_reqs["dev"],
     extras_require=extra_reqs,
+    version="2.4.9",
```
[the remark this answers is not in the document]
vincentsarago: yeah we should remove this

stac_fastapi/extensions/setup.py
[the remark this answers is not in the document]
jonhealy1: The thing is though it's sort of complicated because let's say we put types to 3.0.0 but types has not been release to pypi yet so the tests don't pass.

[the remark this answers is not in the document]
jonhealy1: In sfeos I have to push a new core version manually to pypi before I can push es and os updates with new versions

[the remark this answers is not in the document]
vincentsarago: we won't change the version before the release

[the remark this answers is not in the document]
vincentsarago: this is how we do in titiler and it works fine. 

right now the versioning is quite dangerous because if you have multiple version of the modules you could hit some issues. Because the 3 modules are supper linked to each other, the versions should always be the same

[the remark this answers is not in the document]
jonhealy1: the release is triggered by a github workflow - if we put stac-fastapi.types==3.0.0 in stac-fastapi.api then the tests will fail so how can we merge?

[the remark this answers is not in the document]
jonhealy1: if you don't change the version before the release then the release will point to old versions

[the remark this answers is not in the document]
jonhealy1: I don't think it's dangerous becuase stac-fastapi is run in a container or in the cloud - no one should have old versions installed that won't be re-installed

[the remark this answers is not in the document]
jonhealy1: Does anyone use them separately?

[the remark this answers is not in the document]
jonhealy1: They all have to be on the same version together at the same time pretty much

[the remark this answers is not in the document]
vincentsarago: ok let's discuss this in an issue 👍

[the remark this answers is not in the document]
jonhealy1: See what people think

jonhealy1 (approved):
Nice work!

vincentsarago (approved):
🚀 thanks @thomas-maschler

Notes:
- we should wait for the next stac-pydantic release with https://github.com/stac-utils/stac-pydantic/pull/131
- I've created  a `maint-2.x` branch (https://github.com/stac-utils/stac-fastapi/tree/maint-2.x) if we need to continue updating `old` stac-fastapi

jonhealy1:
@vincentsarago when you say wait do you mean wait to merge?

vincentsarago:
@jonhealy1 sorry I meant, wait for the 3.0 release :-) 

I'll merge this now 🥳

vincentsarago:
🤯 😍 
Screenshot 2024-04-26 at 9 27 17 AM

Screenshot 2024-04-26 at 9 27 15 AM

PR #627, opened 2024-02-12
Updating to stac pydantic 3

rhysrevans3:
**Related Issue(s):**

- #

**Description:**

**PR Checklist:**


thomas-maschler:
@rhysrevans3 looks like we had the same idea
https://github.com/stac-utils/stac-fastapi/pull/625

stac_fastapi/api/stac_fastapi/api/models.py
```diff
@@ -44,31 +43,8 @@ def create_request_model(
...
-                    min_items=field_info.min_items,
-                    max_items=field_info.max_items,
-                    min_length=field_info.min_length,
-                    max_length=field_info.max_length,
-                    regex=field_info.regex,
-                    extra=field_info.extra,
-                )
-                fields[k] = (v.outer_type_, body)
+            for k, field_info in model.model_fields.items():
+                fields[k] = (field_info.annotation, field_info)
```
rhysrevans3, line 47: @thomas-maschler sorry I should have checked before putting in the pull request it looks like most of our changes are the same. I found the `field_info.annotation` would leave out the `Optional` type unless I switched from typing List and Dict to the inbuilt list and dict.

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+* Removing support for Python 3.8
```
vincentsarago, line 5: Why is this needed? pydantic 2 does support python 3.8 https://github.com/pydantic/pydantic/blob/main/pyproject.toml#L65

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8-slim as base
+FROM python:3.9-slim as base
```
vincentsarago, line 1: agree to remove python 3.8 but then we should use python 3.11 IMO

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -67,8 +68,8 @@ class StacApi:
 
     settings: ApiSettings = attr.ib()
     client: Union[AsyncBaseCoreClient, BaseCoreClient] = attr.ib()
-    extensions: List[ApiExtension] = attr.ib(default=attr.Factory(list))
-    exceptions: Dict[Type[Exception], int] = attr.ib(
+    extensions: list[ApiExtension] = attr.ib(default=attr.Factory(list))
+    exceptions: dict[Type[Exception], int] = attr.ib(
```
vincentsarago, line 72: FYI is using `dict` directly instead of `typing.Dict` is the reason to switch to python>=3.9, `typing.Dict` is just an alias of `dict` (https://github.com/python/cpython/blob/3.9/Lib/typing.py#L1743C15-L1743C19) so it's fine to use `typing.Dict` IMO 

https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict

vincentsarago: same goes for List

rhysrevans3: @vincentsarago I found that during the `create_request_model` in models.py. The `field_info.annotation` would leave out the `Optional` part of the type. The only solution I found was to switch from typing List and Dict to the generic list and dict. There maybe a better solution to this?

lossyrob:
Is there a reason to keep both this and #625 open? Based on a quick review, looks like #625 has less changes - and by the author of the stac-pydantic upgrade (thank you @thomas-maschler !). If we can close this and focus the review on #625 that will be easier on reviewers. @rhysrevans3 anything in here related to the pydantic upgrade that deem a must-have that #625 doesn't already contain? Otherwise consolidating on #625 and getting your review on it would be a good path to get this unblocked IMO. Thanks for this PR, unfortunate instance of duplication of work!

Issue #581, opened 2023-06-21
Allowing to fix the request base URL


constantinius:
The base URL in links is built using the Starlette Request `base_url` property. But this value can be misleading, especially when the app is used behind a proxy.

It would help if the base URL can be configured somehow.

betolink:
I was trying to deploy `stac-fastapi` using traefik so that when I hit 
`myurl.com/stac/` the links in the catalog contain the "/stac" prefix. Related to #223 and #581 

I think it would be useful if we could use an environment variable to configure the `--root-path` instead of manually having to modify the fastapi command.

vincentsarago:
https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/config.py#L34

PR #636, opened 2024-02-17
Proxy Friendly Setting

zachcoleman:
**Related Issue(s):**

- #581 

**Description:**

Create a setting to help make it more proxy friendly. 

**PR Checklist:**

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/config.py
+++ b/stac_fastapi/types/stac_fastapi/types/config.py
@@ -29,6 +29,7 @@ class ApiSettings(BaseSettings):
 
     openapi_url: str = "/api"
     docs_url: str = "/api.html"
+    base_url: Optional[str] = None
 
     class Config:
         """Model config (https://pydantic-docs.helpmanual.io/usage/model_config/)."""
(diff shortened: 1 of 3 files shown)


zachcoleman:
I want to be clear that while I think the proxy middleware works, the experience I had was that scheme is not well persisted and that a proxy accepting https forwarding them as http was not being reflected in the links dynamically configured based on base_url.

vincentsarago:
going back to the issue 

> The base URL in links is built using the Starlette Request base_url property. But this value can be misleading, especially when the app is used behind a proxy.

I find this a bit questionable because it's usually a configuration issue. In the different project I managed we've been careful not to add any `hack` around base_url and request's url construction (https://github.com/developmentseed/titiler/discussions/342) 🤔

zachcoleman:
> going back to the issue
> 
> > The base URL in links is built using the Starlette Request base_url property. But this value can be misleading, especially when the app is used behind a proxy.
> 
> I find this a bit questionable because it's usually a configuration issue. In the different project I managed we've been careful not to add any `hack` around base_url and request's url construction ([developmentseed/titiler#342](https://github.com/developmentseed/titiler/discussions/342)) 🤔

I 100% agree this is a hack and will close. 

I do want to point out that this is an issue where the scheme `http` is swapped in for `https`. We have this currently in as a middleware instead of this hack. Using:
```python
class ForwardedSchemeMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
        scheme = request.headers.get("x-forwarded-proto")
        if scheme:
            request.scope["scheme"] = scheme
        response = await call_next(request)
        return response
 ```

PR #639, opened 2024-03-11
Configure the landing page id, description, etc. via env vars

m-mohr:
**Related Issue(s):**

- None?

**Description:**

Allow an easy way to configure the landing page id, description, title and version via env variables.
Maybe there's another way to do it properly, but it was not documented and on Gitter no one seemed to know a way wither.
So trying to enable a way to set the information via environment variable. 

> For the landing page, you can set the API title, description and version using environment variables.
> 
> - `STAC_FASTAPI_VERSION` (string) is the version number of your API instance (this is not the STAC version).
> - `STAC FASTAPI_TITLE` (string) should be a self-explanatory title for your API.
> - `STAC FASTAPI_DESCRIPTION` (string) should be a good description for your API. It can contain CommonMark.
> - `STAC_FASTAPI_ID` (string) is a unique identifier for your API.

**PR Checklist:**


jamesfisher-gis:
Should this be defined here, or in the implementations when they call the `StacApi` class?

See [this PR](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/issues/206)

m-mohr:
I believe it's fine here as it's just the fallback for when it's not passed from the implementations, isn't it?

jamesfisher-gis:
> I believe it's fine here as it's just the fallback for when it's not passed from the implementations. It would fail if the implementatons pass a default value, too.
> 
Actually, none of the implementations include a title or description when calling the `StacApi` class.

This would fail if any `title` or `description` string is passed to the `StacApi` class. Rather, it would quietly use either the environment variable or default value and ignore the user-defined value.

example:
```python
api = StacApi(
    title="My API title that is not an env var",
    description="My API description that is not an env var",
    settings=settings,
    extensions=extensions,
    client=CoreClient(session=session, post_request_model=post_request_model),
    search_get_request_model=create_get_request_model(extensions),
    search_post_request_model=post_request_model,
)
```
The title and description here would both be the default, "stac-fastapi". Could cause confusion.

> Also, I believe that this is something that is not backend specific and as such should in general be in the common implementation, otherwise the behavior is different depending on tha backend. That is not logical to me.
> 
Agreed. But from a code readability perspective I think it is better to have the title and description defined when calling the `StacApi` class in the implementation and leave the `StacApi` class to impose the baseline default ("stac-fastapi").

> If we go forward with [stac-utils/stac-fastapi-elasticsearch-opensearch#207](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/pull/207), we should align the variable names @jamesfisher-gis

Yup. Regardless I changed the env vars here to match yours.

m-mohr:
> This would fail if any `title` or `description` string is passed to the `StacApi` class. Rather, it would quietly use either the environment variable or default value and ignore the user-defined value.

Why is that? I guess then I don't understand how attrs works.

jamesfisher-gis:
> > This would fail if any `title` or `description` string is passed to the `StacApi` class. Rather, it would quietly use either the environment variable or default value and ignore the user-defined value.
> 
> Why is that? I guess then I don't understand how attrs works.
> 
> 

No, you are right. I forgot that the get env was in the default argument.

gadomski (approved):

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -83,10 +84,12 @@ class StacApi:
...
     router: APIRouter = attr.ib(default=attr.Factory(APIRouter))
-    title: str = attr.ib(default="stac-fastapi")
-    api_version: str = attr.ib(default="0.1")
+    title: str = attr.ib(default=os.getenv("STAC_FASTAPI_TITLE", "stac-fastapi"))
+    api_version: str = attr.ib(default=os.getenv("STAC_FASTAPI_VERSION", "0.1"))
     stac_version: str = attr.ib(default=STAC_VERSION)
-    description: str = attr.ib(default="stac-fastapi")
+    description: str = attr.ib(
+        default=os.getenv("STAC_FASTAPI_DESCRIPTION", "stac-fastapi")
+    )
```
vincentsarago, line 92: I'm not quite sure to understand why we set this at the router level and not at the app level?

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -83,10 +84,12 @@ class StacApi:
         converter=update_openapi,
     )
     router: APIRouter = attr.ib(default=attr.Factory(APIRouter))
-    title: str = attr.ib(default="stac-fastapi")
-    api_version: str = attr.ib(default="0.1")
+    title: str = attr.ib(default=os.getenv("STAC_FASTAPI_TITLE", "stac-fastapi"))
```
vincentsarago, line 87: also I think we usually use pydantic settings to take care of the env variable (enable the use of `.env` files too). I think it will be better to do the same here

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -253,9 +254,11 @@ class LandingPageMixin(abc.ABC):
...
 
     stac_version: str = attr.ib(default=STAC_VERSION)
-    landing_page_id: str = attr.ib(default="stac-fastapi")
-    title: str = attr.ib(default="stac-fastapi")
-    description: str = attr.ib(default="stac-fastapi")
+    landing_page_id: str = attr.ib(default=os.getenv("STAC_FASTAPI_ID", "0.1"))
+    title: str = attr.ib(default=os.getenv("STAC_FASTAPI_TITLE", "stac-fastapi"))
+    description: str = attr.ib(
+        default=os.getenv("STAC_FASTAPI_DESCRIPTION", "stac-fastapi")
+    )
```
vincentsarago, line 261: To me it feels weird that we have to set this here again but it's not your fault 😓

stac_fastapi/api/stac_fastapi/api/app.py
m-mohr: I guess that's something to ask the initial authors?

stac_fastapi/api/stac_fastapi/api/app.py
vincentsarago: no worry, I think we will use the solution proposed in #499

m-mohr: Waiting for 3.0.0 doesn't sound ideal to me...

jonhealy1:
We are actually already using this in stac-fastapi-elasticsearch-opensearch. https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/pull/207

jonhealy1:
Could we add this to v2.5.0 and then leading up to v3.0.0 try to reconcile this with #499?

docs/tips-and-tricks.md
```diff
@@ -31,3 +31,12 @@ from stac_fastapi.extensions.core.context import ContextExtension
...
 and then edit the `api = StacApi(...` call to add `ContextExtension()` to the list given as the `extensions` parameter.
+
+## Set API title, description and version
+
+For the landing page, you can set the API title, description and version using environment variables.
+
+- `STAC_FASTAPI_VERSION` (string) is the version number of your API instance (this is not the STAC version).
+- `STAC FASTAPI_TITLE` (string) should be a self-explanatory title for your API.
+- `STAC FASTAPI_DESCRIPTION` (string) should be a good description for your API. It can contain CommonMark.
+- `STAC_FASTAPI_ID` (string) is a unique identifier for your API.
```
vincentsarago, line 42: @m-mohr what's the difference between `STAC_FASTAPI_VERSION` and `STAC_FASTAPI_ID` ?

m-mohr: One sets the id field in the landing page, the other set's the version of STAC fastapi afaik.

vincentsarago: but why would you want those to be different?

vincentsarago: 🤔 ok I think I understand, IMO `STAC_FASTAPI_ID` should be renamed `STAC_FASTAPI_LANDING_ID` to be a bit more verbose. I'm also bit curious how ofter people will want to set this via env variable 🤷

m-mohr: Why should it be kept stac-fastapi? Having all APIs expose the same id seems like a bad practice.
Maybe the names should actually be:

- STAC_CATALOG_ID
- STAC_CATALOG_TITLE
- STAC_CATALOG_DESCRIPTION
- STAC_FASTAPI_INSTANCE_VERSION

?

vincentsarago: 🤔 I'm not sure I quite get this. if you're talking about the env prefix `STAC_FASTAPI`, to me it makes sense because you want to setup something within the stac-fastapi application! This way you're trying not to conflict with other application. 

In term of variable, I think it's a big mess because we're using the variable to set information in the OpenAPI document and the Landing page which are both totally different 🤷

m-mohr: Okay, fine for me. I wouldn't add the `_LANDING` myself because it is an implementation detail. Imaging OpenAPI 4 having an id field in the info section, then you would want to remove the LANDING_ again. But as long as there's a way to configure it, I'm fine with whatever is done ;-)

docs/src/tips-and-tricks.md
```diff
@@ -31,3 +31,12 @@ from stac_fastapi.extensions.core.context import ContextExtension
...
 and then edit the `api = StacApi(...` call to add `ContextExtension()` to the list given as the `extensions` parameter.
+
+## Set API title, description and version
+
+For the landing page, you can set the API title, description and version using environment variables.
+
+- `STAC_FASTAPI_VERSION` (string) is the version number of your API instance (this is not the STAC version).
+- `STAC FASTAPI_TITLE` (string) should be a self-explanatory title for your API.
+- `STAC FASTAPI_DESCRIPTION` (string) should be a good description for your API. It can contain CommonMark.
+- `STAC_FASTAPI_ID` (string) is a unique identifier for your API.
```
vincentsarago, line 42:
```suggestion
- `STAC_FASTAPI_LANDING_ID` (string) is a unique identifier for your API.
```

PR #654, opened 2024-04-10
add python 3.12 support and some docker cleanup

vincentsarago:
This PR does:
- add python 3.12 support 
- removes unused docker service in CI
- other: `pip` -> `python -m pip` (this is to make sure it use the correct pip or pytest module when working with virtual environment)


jonhealy1 (approved):
Looks good

PR #657, opened 2024-04-11
use pydantic settings

vincentsarago:
proposed changed for #639 

This PR is moving the env configuration into `ApiSettings`

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/config.py
+++ b/stac_fastapi/types/stac_fastapi/types/config.py
@@ -25,7 +25,7 @@ class ApiSettings(BaseSettings):
     stac_fastapi_title: str = "stac-fastapi"
     stac_fastapi_description: str = "stac-fastapi"
     stac_fastapi_version: str = "0.1"
-    stac_fastapi_id: str = "0.1"
+    stac_fastapi_landing_id: str = "stac-fastapi"
 
     app_host: str = "0.0.0.0"
     app_port: int = 8000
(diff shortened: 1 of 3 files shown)


jonhealy1 (approved):
LGTM

stac_fastapi/types/stac_fastapi/types/config.py
```diff
@@ -22,6 +22,11 @@ class ApiSettings(BaseSettings):
     # `pydantic.BaseSettings` instead
     default_includes: Optional[Set[str]] = None
 
+    stac_fastapi_title: str = "stac-fastapi"
+    stac_fastapi_description: str = "stac-fastapi"
+    stac_fastapi_version: str = "0.1"
+    stac_fastapi_landing_id: str = "stac-fastapi"
```
vincentsarago, line 28: as discussed in #639 I think it makes more sense for this variable to be named `landing_id`

vincentsarago, line 28: I've also reverted to `stac-fastapi` default as it was before 

https://github.com/stac-utils/stac-fastapi/blob/315cfaefef2ab2773ae087228439ba0371fb7372/stac_fastapi/types/stac_fastapi/types/core.py#L258

vincentsarago:
#657 @jonhealy1 could you check with the latest change 🙏 

@m-mohr I'll wait for your 👍

stac_fastapi/types/stac_fastapi/types/config.py
jonhealy1: Looks good to me. The version is the version number of the User's stac_fastapi, not this version ie. soon to be 2.5.0

PR #670, opened 2024-04-23
return 400 for datetime errors

jonhealy1:
**Related Issue(s):**

- #

**Description:**

**PR Checklist:**

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/rfc3339.py
+++ b/stac_fastapi/types/stac_fastapi/types/rfc3339.py
@@ -73,10 +73,13 @@ def validate_interval_format(values: list) -> None:
         values (list): A list of strings split by '/' from the interval string.
 
     Raises:
-        ValueError: If the interval string contains more than one forward slash.
+        HTTPException: If the interval string contains more than one forward slash.
     """
     if len(values) > 2:
-        raise ValueError("Interval string contains more than one forward slash.")
+        raise HTTPException(
+            status_code=400,
+            detail="Interval string contains more than one forward slash.",
+        )
 
 
 def str_to_interval(interval: Optional[str]) -> Optional[DateTimeType]:
--- a/stac_fastapi/types/stac_fastapi/types/rfc3339.py
+++ b/stac_fastapi/types/stac_fastapi/types/rfc3339.py
@@ -4,6 +4,7 @@ from datetime import datetime, timezone
 from typing import Optional, Tuple, Union
 
 import iso8601
+from fastapi import HTTPException
 from pystac.utils import datetime_to_str
 
 RFC33339_PATTERN = (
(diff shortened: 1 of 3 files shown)


stac_fastapi/types/stac_fastapi/types/rfc3339.py
```diff
@@ -45,53 +46,88 @@ def rfc3339_str_to_datetime(s: str) -> datetime:
...
+        date_str (str): A string representing the date in RFC3339 format.
+
+    Returns:
+        datetime: A datetime object parsed from the date_str.
+
+    Raises:
+        ValueError: If the date_str is empty or contains the placeholder '..'.
+    """
+    if ".." in date_str or not date_str:
+        raise ValueError("Invalid date format.")
```
vincentsarago, line 63: why not raise 400 here too ?

stac_fastapi/types/stac_fastapi/types/rfc3339.py
```diff
@@ -45,53 +46,88 @@ def rfc3339_str_to_datetime(s: str) -> datetime:
...
+        raise HTTPException(status_code=400, detail="Empty interval string is invalid.")
 
     values = interval.split("/")
-    if len(values) == 1:
-        # Single date for == date case
-        return rfc3339_str_to_datetime(values[0])
-    elif len(values) > 2:
-        raise ValueError(
-            f"Interval string '{interval}' contains more than one forward slash."
+    validate_interval_format(values)
```
vincentsarago, line 109: kinda weird to me to create a function just for this

stac_fastapi/types/stac_fastapi/types/rfc3339.py
jonhealy1: the only place this code is accessed is down below via parse_single_date - 
```
try:
        start = parse_single_date(values[0]) if values[0] not in ["..", ""] else None
        end = (
            parse_single_date(values[1])
            if len(values) > 1 and values[1] not in ["..", ""]
            else None
        )
    except ValueError as e:
        raise HTTPException(status_code=400, detail=str(e))

File: stac_fastapi/types/stac_fastapi/types/rfc3339.py (hunk 2 above)
Comment (jonhealy1) [reply stac_fastapi/types/stac_fastapi/types/rfc3339.py:109]: It's true - I was getting a complication error before with linting - I think I went too far in breaking things up

File: stac_fastapi/types/stac_fastapi/types/rfc3339.py [hunk 3]
```diff
@@ -45,53 +46,88 @@ def rfc3339_str_to_datetime(s: str) -> datetime:
...
         )
-
-    start = None
-    end = None
-    if values[0] not in ["..", ""]:
-        start = rfc3339_str_to_datetime(values[0])
-    if values[1] not in ["..", ""]:
-        end = rfc3339_str_to_datetime(values[1])
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e))
```
vincentsarago, line 119: I would personally raise HTTPException witting the `parse_single_date` function and remove the `try .. except` block .

stac_fastapi/types/stac_fastapi/types/rfc3339.py
vincentsarago: just add `#noqa: C901` on top of the function to ignore those linting issue

jonhealy1: ok

stac_fastapi/types/stac_fastapi/types/rfc3339.py
jonhealy1: I think this is hard because the try except catches errors that filter from this function --- rfc3339_str_to_datetime via parse_single_date

jonhealy1: We could put the try except in parse_single_date?

vincentsarago: the other solution would be to change https://github.com/stac-utils/stac-fastapi/blob/73402d9b58c0291f18d3b128a1d417ccaf4bd1aa/stac_fastapi/types/stac_fastapi/types/rfc3339.py#L43C15-L43C25 and raise a specific error `InvalidDatetime` and then catch this in https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/api/stac_fastapi/api/errors.py#L23-L30

jonhealy1: Things seem to pass the re.match and then error out here `return iso8601.parse_date(date_str)` hence the try except: https://github.com/stac-utils/stac-fastapi/blob/73402d9b58c0291f18d3b128a1d417ccaf4bd1aa/stac_fastapi/types/stac_fastapi/types/rfc3339.py#L40-L46

jonhealy1: I can't remember what exactly now

jonhealy1: Not everything

vincentsarago (approved):

vincentsarago:
FYI: this might gets revisited after https://github.com/stac-utils/stac-pydantic/pull/135

PR #673, opened 2024-04-24
Patch/fix doc urls in landing

vincentsarago:
ref https://github.com/stac-utils/stac-fastapi-pgstac/pull/101/files#r1576403075

This PR fix a bug introduced in #635 

I've added tests (taken mostly from #625 )

PR #675, opened 2024-04-25
fix AsyncBaseCoreClient urls

vincentsarago:
same as #673 but for Async 🤦 

I think we can simply do a `post` release for this we don't have to got with 2.5.6


jonhealy1 (approved):

jonhealy1:
Do we update the version numbers for pypi? I guess we have to

PR #679, opened 2024-04-26
use Collection Pydantic model in PutCollection transaction

vincentsarago:
Not sure why this was not in #625 

Note: we should add a note in the changelog that transaction endpoint now have Pydantic Model instead of TypedDict as input (https://github.com/stac-utils/stac-fastapi-pgstac/pull/108#issuecomment-2080119937)

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
@@ -10,7 +10,6 @@ from starlette.responses import JSONResponse, Response
 
 from stac_fastapi.api.models import CollectionUri, ItemUri
 from stac_fastapi.api.routes import create_async_endpoint
-from stac_fastapi.types import stac
 from stac_fastapi.types.config import ApiSettings
 from stac_fastapi.types.core import AsyncBaseTransactionsClient, BaseTransactionsClient
 from stac_fastapi.types.extension import ApiExtension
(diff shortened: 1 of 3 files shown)


stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -38,7 +38,7 @@ def create_item(
         collection_id: str,
         item: Union[Item, ItemCollection],
         **kwargs,
-    ) -> Optional[Union[Item, Response, None]]:
+    ) -> Optional[Union[stac.Item, Response, None]]:
```
vincentsarago, line 41: To Be Discussed! 

It seems weird to return a Pydantic Object, while non-transaction endpoints return `stac types` (typedDict) object

jonhealy1: I would agree - we don't want to return pydantic objects

jonhealy1 (approved):

stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
```diff
@@ -34,7 +33,7 @@ class PutItem(ItemUri):
 class PutCollection(CollectionUri):
     """Update Collection."""
 
-    collection: stac.Collection = attr.ib(default=Body(None))
+    collection: Collection = attr.ib(default=Body(None))
```
vincentsarago, line 36: @jonhealy1 I'm not sure why but when we use this class (attr) we don't get the validation/serialization 😬 

```python
from stac_fastapi.extensions.core.transaction import PutCollection

>>> PutCollection(collection_id='yo', collection={})
PutCollection(collection_id='yo', collection={})

>>> print(type(PutCollection(collection_id='yo', collection={}).collection))
<class 'dict'>
```

you may have seen this also in the elastic search backend

Issue #615, opened 2023-11-03
Add support for the STAC API Collection Search Extension


juliayun23:
https://github.com/stac-api-extensions/collection-search

I noticed that [stac-browser v3.1.0](https://github.com/radiantearth/stac-browser/releases/tag/v3.1.0) has added support for Collection Search. We are using `stac-browser `on top of `stac-fastapi`. 

Just wondering whether there's any plan of adding the support for Collection Search Extension in `stac-fastapi`? Thanks.

vincentsarago:
partially done in #736
post extension proposed in #739

PR #681, opened 2024-05-03
Adding Collection Search Extension

tjellicoe-tpzuk:
**Related Issue(s):** #615 

**Description:** Adding Collection Search Extension as defined [here](https://github.com/stac-api-extensions/collection-search)

**PR Checklist:**


stac_fastapi/extensions/stac_fastapi/extensions/core/collectionSearch/collectionSearch.py
```diff
@@ -0,0 +1,108 @@
...
+    SIMPLE_QUERY = "http://www.opengis.net/spec/ogcapi-common-2/1.0/conf/simple-query"
+
+@attr.s
+class CollectionSearchExtension(ApiExtension):
+    """CollectionSearch Extension.
+
+    The collection search extension adds two endpoints which allow searching of 
+    collections via GET and POST:
+        GET /collection-search
+        POST /collection-search
```
vincentsarago, line 37: we should add a note that the extension seems to mention `GET /collections` and `POST /collections` but due to possible conflict with the `transaction` extension we should to use `/collection-search`

also wonder if we should use `/collections-search` or `/search/collections` 🤷‍♂️

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -175,7 +177,7 @@ def register_get_item(self):
         """
         self.router.add_api_route(
             name="Get Item",
-            path="/collections/{collection_id}/items/{item_id}",
+            path="/catalogs/{catalog_id}collections/{collection_id}/items/{item_id}",
```
vincentsarago, line 180: 😬 maybe this change wasn't made on purpose?

m-mohr (changes requested):
This doesn't implement https://github.com/stac-api-extensions/collection-search

The extension doesn't define a POST /collections endpoint, only GET /collections. The endpoint names are wrong.

> All functionality in OGC API - Records - Part 1: Local Resource Catalogue is only defined for the GET method (i.e. GET /collections). Note: STAC may add behavior for POST /collections in the future, but due to a potential conflict with the Transaction Extension, specific rules for content negotiation might be required.

tjellicoe-tpzuk:
Thanks for your comments @vincentsarago and @m-mohr, this PR has become mixed up with subsequent commits, I will raise a new one only containing the updates for the collection search extension - see  draft PR #693

Issue #648, opened 2024-04-04
Introduce a tool to bump all version strings across project before release


thomas-maschler:
Introduce a tool and allow to automatically bump version via CI before a release based on the release tag.

Before a release, we always need to manually update version strings in multiple places across the project. This is very error-prone.
There are many tools out there for example `bump2version`
https://github.com/developmentseed/titiler/blob/main/.bumpversion.cfg

PR #685, opened 2024-05-06
Release/v3.0.0a0

vincentsarago:
Q: `3.0.0.alpha0` or `3.0.0a0` ?

also takes care of #648 

```
bump2version major --new-version 3.0.0.alpha0 --no-tag 
```


jonhealy1:
Should we put the bumpversion command example in the CONTRIBUTING.md file?

vincentsarago:
`pystac` uses `3.0.0a0` notation so I guess we should do it as well

PR #691, opened 2024-05-14
update conformance and spec urls

vincentsarago:
This PR updates the conformance and spec URLs for the transaction extension.

The Transaction extension is now split in two extensions:
- https://github.com/stac-api-extensions/transaction
- https://github.com/stac-api-extensions/collection-transaction

note: I first thought about splitting the TransactionExtension class in two but it doesn't make much sense if we keep only one `Client` (BaseTransactionsClient). If we really wanted this we would need to also split the `BaseTransactionsClient` and thus have a lot of duplicated code.

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
@@ -49,17 +49,20 @@ class TransactionExtension(ApiExtension):
         PUT /collections/{collection_id}/items
         DELETE /collections/{collection_id}/items
 
-    https://github.com/radiantearth/stac-api-spec/blob/master/ogcapi-features/extensions/transaction/README.md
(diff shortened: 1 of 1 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
```diff
@@ -132,6 +135,11 @@ def register_delete_item(self):
             endpoint=create_async_endpoint(self.client.delete_item, ItemUri),
         )
 
+    def register_patch_item(self):
+        """Register patch item endpoint (PATCH
+        /collections/{collection_id}/items/{item_id})."""
+        raise NotImplementedError
```
vincentsarago, line 141: Make it a bit more clear that the `Patch` method are not implemented by purpose!

jonhealy1 (approved):

Issue #688, opened 2024-05-07
How to tell if search is for single datetime? (Should `str_to_interval` return a single value for that?)


avbentem:
I'd guess this is a bug, but maybe not, so a question first:

The [STAC API - Item Search specs define](https://github.com/radiantearth/stac-api-spec/tree/main/item-search#query-parameter-table):

> `datetime`
>
> Single date+time, or a range ('/' separator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). Use double dots `..` for open date ranges.

**If only a single date+time is provided, I'd guess this implies one searches for an exact match?**

Indeed, [`DateTimeType` allows for a single value too](https://github.com/stac-utils/stac-fastapi/blob/add05de82f745a717b674ada796db0e9f7153e27/stac_fastapi/types/stac_fastapi/types/rfc3339.py#L16-L21):

```python
DateTimeType = Union[
    datetime,
    Tuple[datetime, datetime],
    Tuple[datetime, None],
    Tuple[None, datetime],
]
```

All good so far. However, today's `str_to_interval` [always returns a tuple](https://github.com/stac-utils/stac-fastapi/blob/add05de82f745a717b674ada796db0e9f7153e27/stac_fastapi/types/stac_fastapi/types/rfc3339.py#L117). And that function [is used as the converter for the search request](https://github.com/stac-utils/stac-fastapi/blob/add05de82f745a717b674ada796db0e9f7153e27/stac_fastapi/types/stac_fastapi/types/search.py#L63):

```python
class BaseSearchGetRequest(APIRequest):
    ...
    datetime: Optional[DateTimeType] = attr.ib(default=None, converter=str_to_interval)
   ...
```

So when implementing [get_search](https://github.com/stac-utils/stac-fastapi/blob/add05de82f745a717b674ada796db0e9f7153e27/stac_fastapi/types/stac_fastapi/types/core.py#L454-L464) how would one know if a single date+time was requested, or a range with an open end?

[The tests](https://github.com/stac-utils/stac-fastapi/blob/add05de82f745a717b674ada796db0e9f7153e27/stac_fastapi/types/tests/test_rfc3339.py#L88-L99) only check intervals, so are not giving me the definitive answer. But I'm guessing this is an oversight?

(Of course I can provide a PR if this is indeed a bug.)

jonhealy1:
[...]

PR #692, opened 2024-05-14
Make `str_to_interval` not return a tuple for single-value input

avbentem:
**Related Issue(s):**

- #688

**Description:**

This is a simple fix to return a single `datetime` object if a single value (rather than an interval) is given. This is already supported by `DateTimeType`:

https://github.com/stac-utils/stac-fastapi/blob/f815c236cde78a3442fe18c103f43ee1db668872/stac_fastapi/types/stac_fastapi/types/rfc3339.py#L15-L20

It then also changes the value of the `datetime` argument [as passed to `get_search`](https://github.com/stac-utils/stac-fastapi/blob/f815c236cde78a3442fe18c103f43ee1db668872/stac_fastapi/types/stac_fastapi/types/core.py#L454-L464), which so far is passed a `tuple` for a single-value datetime as well.

**PR Checklist:**

- [x] `pre-commit` hooks pass locally
- [x] Tests pass (run `make test`)
- [x] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/stac-fastapi/blob/main/CHANGES.md).

PR #697, opened 2024-05-21
Feature/update stac pydantic3.1.0

vincentsarago:
update to https://github.com/stac-utils/stac-pydantic/releases/tag/3.1.0

also take care of https://github.com/stac-utils/stac-fastapi/issues/688

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/rfc3339.py
+++ b/stac_fastapi/types/stac_fastapi/types/rfc3339.py
@@ -122,6 +122,9 @@ def str_to_interval(interval: Optional[str]) -> Optional[DateTimeType]:
             detail="Interval string contains more than one forward slash.",
         )
 
+    if len(values) == 1:
+        values = [values[0], values[0]]
+
     try:
         start = parse_single_date(values[0]) if values[0] not in ["..", ""] else None
         end = (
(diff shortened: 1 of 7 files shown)


stac_fastapi/types/stac_fastapi/types/rfc3339.py
```diff
@@ -122,6 +122,9 @@ def str_to_interval(interval: Optional[str]) -> Optional[DateTimeType]:
             detail="Interval string contains more than one forward slash.",
         )
 
+    if len(values) == 1:
+        values = [values[0], values[0]]
+
```
vincentsarago, line 127: This is to make sure GET (attr class) and POST (stac-pydantic model) requests will return the same results

jonhealy1 (approved):

stac_fastapi/types/stac_fastapi/types/rfc3339.py
```diff
@@ -122,6 +122,9 @@ def str_to_interval(interval: Optional[str]) -> Optional[DateTimeType]:
             detail="Interval string contains more than one forward slash.",
         )
 
+    if len(values) == 1:
+        values = [values[0], values[0]]
```
vincentsarago, line 126: well I think this wasn't really what was need 😓 
I think https://github.com/stac-utils/stac-fastapi/pull/692 does it better

Issue #700, opened 2024-05-29
Ensure Optional `query` Field Defaults to None for Pydantic V2 Compatibility


joshimai:
I've identified an issue with the Pydantic model definition in `QueryExtensionPostRequest` [stac_fastapi.extensions.core.query.request] (https://github.com/stac-utils/stac-fastapi/blob/1c3546a5b449edebc7587dab289f73224847c63e/stac_fastapi/extensions/stac_fastapi/extensions/core/query/request.py#L21). 
The query field lacks a default value, making it required under Pydantic V2 changes.  In Pydantic V2, fields declared as Optional must explicitly default to None to be considered optional and nullable. This change would prevent potential errors from fields being unexpectedly required and aligns with Pydantic V2's [migration guidelines](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields).

PR #701, opened 2024-05-31
set default value for Query attribute for `QueryExtensionPostRequest` model

vincentsarago:
#700


stac_fastapi/extensions/tests/test_query.py
```diff
@@ -0,0 +1,95 @@
+import json
```
vincentsarago, line 1: added some tests and confirmed that without the change it was failing!

jonhealy1 (approved):
Looks good

PR #705, opened 2024-06-10
Allow default route dependencies

rhysrevans3:
**Description:**
Allows for `*` to be used for the `path` or `method` variables in a route dependencies' scope to match all allowable routes.

**PR Checklist:**


stac_fastapi/api/stac_fastapi/api/routes.py
```diff
@@ -100,28 +101,39 @@ def add_route_dependencies(
...
-                    ),
-                )
-
-            # Register dependencies directly on route so that they aren't ignored if
-            # the routes are later associated with an app (e.g.
-            # app.include_router(router))
-            # https://github.com/tiangolo/fastapi/blob/58ab733f19846b4875c5b79bfb1f4d1cb7f4823f/fastapi/applications.py#L337-L360
-            # https://github.com/tiangolo/fastapi/blob/58ab733f19846b4875c5b79bfb1f4d1cb7f4823f/fastapi/routing.py#L677-L678
-            route.dependencies.extend(dependencies)
+            if hasattr(route, "dependant"):
```
rhysrevans3, line 123: This is needed for paths `/api`, `/api.html', `/docs/oauth2-redirect` as they don't have the dependant attribute.

stac_fastapi/api/stac_fastapi/api/routes.py
```diff
@@ -100,28 +101,38 @@ def add_route_dependencies(
...
     """
     for scope in scopes:
+        _scope = copy.deepcopy(scope)
         for route in routes:
-            match, _ = route.matches({"type": "http", **scope})
+            if scope["path"] == "*":
+                _scope["path"] = route.path
+
+            if scope["method"] == "*":
+                _scope["method"] = list(route.methods)[0]
```
jonhealy1, line 116: Is this correct or should we apply dependencies to each method individually?

rhysrevans3: I believe so, the method is only used to get a full match with the route [here](https://github.com/stac-utils/stac-fastapi/pull/705/files#diff-3f814cfc24182540359cf16e8a3cedff38803d2feb6e489f7fc83b2e46758287R118-R120). The dependency is then applied to the route so is not specific to the method.

vincentsarago:
@alukach would you be able to review this one? 🙏

jonhealy1 (approved):
LGTM nice work!

stac_fastapi/api/stac_fastapi/api/routes.py
alukach: If we just use a guard here, we can avoid indenting the remaining code.
```suggestion
            # Ignore paths without dependants, e.g. /api, /api.html, /docs/oauth2-redirect
            if not hasattr(route, "dependant"):
                continue
```

alukach (approved):
This is a nice convenience feature.  I think it makes sense. At first I thought a user should instead just do something like:

```py
routes.add_route_dependencies(
    app.router.routes,
    [
        {"path": route.path, "method": method, "type": "http"}
        for route in app.router.routes
        for method in route.methods
    ],
    [Depends(auth.basic_auth)],
)
```

However, the value of being able to something like this seems really enticing:

```py
routes.add_route_dependencies(
    app.router.routes,
    [
        {"path": "*", "method": method, "type": "http"}
        for method in ("POST", "PUT", "DELETE")
    ],
    [Depends(auth.basic_auth)],
)
```

Left a minor code style suggestion, but 👍 💯

vincentsarago (approved):

Issue #642, opened 2024-03-25
Simple search returns items onlly with properties.datetime field


iliion:
Dear all.
 I have already opened an issue in `stac-fastapi-elasticsearch`  repository and I will report it here as well. See [issue](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/issues/217#issue-2202761073).

---

Simple search `http://localhost:8000/search?limit=10` returns items onlly with properties.datetime field

In particular in line
https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/blob/987f0b924b45e3f63302758b6e63d38ba504964f/stac_fastapi/core/stac_fastapi/core/core.py#L593C12-L593C64

`filter_kwargs = search_request.fields.filter_fields`

filter_fields is equal to
```
{'exclude': {}, 'include': {'assets': Ellipsis, 'bbox': Ellipsis, 'collection': Ellipsis, 'geometry': Ellipsis, 'id': Ellipsis, 'links': Ellipsis, 'properties': {'datetime'}, 'stac_version': Ellipsis, 'type': Ellipsis}}
```

As a response i get the FeatureCollection
```
{
  "type": "FeatureCollection",
  "features": [ {
    "type": "Feature",
    "properties": {
      "datetime": "2015-07-04T10:10:06.027000+00:00"
    },
    . . .
  ]
  . . .
}
```

Shouldnt the filter_kwargs` be equal to 
```
{'exclude': {}, 'include': {'assets': Ellipsis, 'bbox': Ellipsis, 'collection': Ellipsis, 'geometry': Ellipsis, 'id': Ellipsis, 'links': Ellipsis, 'properties': Ellipsis, 'stac_version': Ellipsis, 'type': Ellipsis}}
```

---

FYI.
The item has the following properties

```
"properties": {
  "sat:relative_orbit": 22,
  "start_datetime": "2015-07-04T10:10:06.027Z",
  "end_datetime": "2015-07-04T10:10:06.027Z",
  "processing:facility": "EPA_",
  "title": "S2A_MSIL2A_20150704T101006_N0204_R022_T30NYK_20150704T102420.SAFE",
  "platform": "Sentinel-2A",
  "view:sun_elevation": 30.130819756471,
  "datetime": "2015-07-04T10:10:06.027Z",
  "instruments": [],
  "constellation": "sentinel-2",
  "sat:orbit_state": "descending",
  "eo:cloud_cover": 84.456044,
[...]

PR #706, opened 2024-06-12
deprecate Fields default-includes

vincentsarago:
ref: https://github.com/stac-utils/stac-fastapi/issues/642, https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/issues/217

This PR does:
- removes `default_includes` from the APISettings
- removes `default_includes` attribute from `FieldsExtension` (this was in fact creating a circular dependency between the Extension and the APISettings 🤯)
-  add deprecation warning for `PostFieldsExtension.filter_fields` this should be implemented by each backend (as stac-fastapi-pgstac already does)


jonhealy1 (approved):
Nice work

PR #714, opened 2024-06-24
replace attr with dataclass + fastapi.Query() for GET models 

### LINKED ISSUE #638: Aliases in FilterExtensionGetRequest model are not working
`FilterExtensionGetRequest` uses `pydantic.Field` instead of `attr.ib` to define aliases. This does not work and fields `filter_crs` and `filter_lang` are not recognized.

```python
from stac_fastapi.extensions.core.filter.request import FilterExtensionGetRequest
filter = FilterExtensionGetRequest(filter="test", filter_crs="EPSG:4326")

# >>> TypeError: FilterExtensionGetRequest.__init__() got an unexpected keyword argument 'filter_crs'
```
```python
filter = FilterExtensionGetRequest(filter="test", **{"filter-crs":"EPSG:4326"})

# >>> TypeError: FilterExtensionGetRequest.__init__() got an unexpected keyword argument 'filter-crs'
```

When trying to define the fields using `attr.ib` hypen aliases are not excepted
```python
from typing import Optional

import attr

from stac_fastapi.types.search import APIRequest
from stac_fastapi.extensions.core.filter.request import FilterLang

@attr.s
class FilterExtensionGetRequest(APIRequest):
    """Filter extension GET request model."""

    filter: Optional[str] = attr.ib(default=None)
    filter_crs: Optional[str] =  attr.ib(alias="filter-crs", default=None)
    filter_lang: Optional[FilterLang] =  attr.ib(alias="filter-lang", default="cql2-text")

# >>>  File <attrs generated init __main__.FilterExtensionGetRequest>:1
#    def __init__(self, filter=attr_dict['filter'].default, filter-crs=attr_dict['filter_crs'].default, filter-lang=attr_dict['filter_lang'].default):
#                                                                 ^
# SyntaxError: invalid syntax
```

I tried various other approaches including adding an `__init__` method to sanitize fields name with hyphens but that didn't play nice with `create_async_endpoint`.

What was the reason to use `attr` classes over Pydantic classes for GET Request Models?

### LINKED ISSUE #713: FilterExtensionGetRequest invalid fields
https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/request.py#L13-L20

The `filter_crs` and `filter_lang` fields are invalid because `FilterExtensionGetRequest` is an `attrs` class, but it uses `Field` from pydantic.

If I try to init the class and define `filter_lang`, for example.

```python
from typing import Any, Dict, Literal, Optional

import attr
from pydantic import BaseModel, Field

from stac_fastapi.types.search import APIRequest

FilterLang = Literal["cql-json", "cql2-json", "cql2-text"]


@attr.s
class FilterExtensionGetRequest(APIRequest):
    """Filter extension GET request model."""

    filter: Optional[str] = attr.ib(default=None)
    filter_crs: Optional[str] = Field(alias="filter-crs", default=None)
    filter_lang: Optional[FilterLang] = Field(alias="filter-lang", default="cql2-text")


FilterExtensionGetRequest(
    filter="{}",
    filter_crs="",
    filter_lang="cql2-json"
)
# TypeError: init() got an unexpected keyword argument 'filter_crs'
```

the class also does not impose the default values for `filter_crs` and `filter_lang`
```python
FilterExtensionGetRequest()
# FilterExtensionGetRequest(filter=None)
```

I think there are two options:
- Make `filter_crs` and `filter_lang` into `attrs` fields. AFAIK `attrs` does not support applying an alias to fields. 
- Make `FilterExtensionGetRequest` into a pydantic class.

Why do we have GET request models as `attrs` classes and POST request models as `pydantic` classes?

vincentsarago:
closes #713 
closes #638


stac_fastapi/api/stac_fastapi/api/models.py
```diff
@@ -37,11 +38,11 @@ def create_request_model(
 
     mixins = mixins or []
 
-    models = [base_model] + extension_models + mixins
+    models = extension_models + mixins + [base_model]
```
vincentsarago, line 41: ⚠️ if we don't do this then dataclass will raise an error because optional fields could be placed before non optional fields 

I don't really like this because it makes mandator that mixing or extensions have only Optional parameters

jonhealy1:
This is still a wip?

vincentsarago:
@jonhealy1 no, this is ready for review. but I'll merge the other PR first

jonhealy1 (changes requested):
Changelog :) Looks good otherwise

jonhealy1 (approved):

PR #723, opened 2024-06-27
Release/v3.0.0a4

vincentsarago:


jonhealy1 (approved):

vincentsarago:
> Is the performance alert something to be concerned about?

not that much, I would be worry if all the timings would change or if there would be a huge difference

PR #725, opened 2024-07-03
remove FieldsExtension check in StacApi

### LINKED ISSUE #709: remove FieldsExtension dependency in `stac-fastapi.api`
in 👇 
https://github.com/stac-utils/stac-fastapi/blob/51a756d904e4bab245da4a525e961d64bcbf1321/stac_fastapi/api/stac_fastapi/api/app.py#L255-L263

We check if the `FieldsExtension` extension is part of the allowed extension, and then remove output model validation if present (because when using fields, the output Item can be an invalid STAC Item). 

This works OK, but brings multiple issues:
- What if no `fields` parameter is passed and the application is set to validate the items?
- it makes `stac-fastapi.api` depends on `stac-fastapi.extensions`

IMO, this is an implementation issue that `stac-fastapi.api` shouldn't take care.

At the implementation level should could be dealt like:

```python
class BadCoreClient(BaseCoreClient):
    def post_search(
        self, search_request: BaseSearchPostRequest, **kwargs
    ) -> stac.ItemCollection:

        search_results = ...

        if getattr("fields", search_request):
            return GeoJSONResponse(search_results)

        return stac.ItemCollection(**search_results)

    def get_search(
        self,
        collections: Optional[List[str]] = None,
        ids: Optional[List[str]] = None,
        bbox: Optional[List[NumType]] = None,
        intersects: Optional[str] = None,
        datetime: Optional[Union[str, datetime]] = None,
        limit: Optional[int] = 10,
        **kwargs,
    ) -> stac.ItemCollection:

        search_results = ...

        if kwargs.get("fields"):
            return GeoJSONResponse(search_results)

        return stac.ItemCollection(**search_results)
```

☝️ This pseudo code show that when a `fields` parameter is passed we should return the `search_results` directly by returning a `Response` object, which won't be `validated` by fastAPI

vincentsarago:
closes #709 

This is the last **breaking change** that I'm thinking about for 3.0 

With this PR, there are no more circular dependencies with `api` and `extensions` 🥳

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/models.py
+++ b/stac_fastapi/api/stac_fastapi/api/models.py
@@ -1,6 +1,5 @@
 """Api request/response models."""
 
-import importlib.util
 from dataclasses import dataclass, make_dataclass
 from typing import List, Optional, Type, Union
 
@@ -18,6 +17,12 @@ from stac_fastapi.types.search import (
     str2bbox,
     str_to_interval,
 )
+
+try:
+    import orjson  # noqa
+    from fastapi.responses import ORJSONResponse as JSONResponse
+except ImportError:  # pragma: nocover
+    from starlette.responses import JSONResponse
 
 
 def create_request_model(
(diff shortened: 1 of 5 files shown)


stac_fastapi/api/stac_fastapi/api/models.py
```diff
@@ -19,6 +18,12 @@
     str_to_interval,
 )
 
+try:
+    import orjson  # noqa
+    from fastapi.responses import ORJSONResponse as JSONResponse
+except ImportError:  # pragma: nocover
+    from starlette.responses import JSONResponse
+
```
vincentsarago, line 26: This is slightly un-related to this PR goals but I couldn't 🙈 on some lines 🔪

jonhealy1 (approved):
Nice documenting!

PR #729, opened 2024-07-08
move back to attrs

### LINKED ISSUE #728: `make_dataclass` erase the `__post_init__` method
in #714 we changed how we create GET request model from using `attrs` to `dataclass`. This was to support `alias` with `-`.  The downside was that we need to do `validation/conversion` in a `__post_init__` method.

It seems that this wasn't well tested 😢, and now when we create a Model with `extensions` the `__post_init__` is not ran 😭 😭 😭 😭 😭 

```python
from stac_fastapi.api.models import create_get_request_model
from stac_fastapi.extensions.core import FieldsExtension

model = create_get_request_model(extensions=[FieldsExtension()])
model(datetime="yo")


model = create_get_request_model(extensions=[])
model(datetime="yo")

>> HTTPException: 400: Invalid RFC3339 datetime.
```

### where 

https://github.com/stac-utils/stac-fastapi/blob/494e485c5d619d8c800bc845c391dfdced21fb2a/stac_fastapi/api/stac_fastapi/api/models.py#L28-L50

vincentsarago:
closes #728 

well, I found a way to still user `attr` and allow `-` in aliases 🎉

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/fields/request.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/fields/request.py
@@ -1,9 +1,9 @@
 """Request models for the fields extension."""
 
 import warnings
-from dataclasses import dataclass
 from typing import Dict, Optional, Set
 
+import attr
 from fastapi import Query
 from pydantic import BaseModel, Field
 from typing_extensions import Annotated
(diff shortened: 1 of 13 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/core/fields/request.py
```diff
@@ -70,16 +70,11 @@ def filter_fields(self) -> Dict:
...
-@dataclass
+@attr.s
 class FieldsExtensionGetRequest(APIRequest):
     """Additional fields for the GET request."""
 
-    fields: Annotated[Optional[str], Query()] = None
-
-    def __post_init__(self):
-        """convert attributes."""
-        if self.fields:
```
vincentsarago, line 81: 🤦 I think the whole issue came from this.

I should have added `super().__post_init__()` here (maybe) 

the attrs solution is definitely better anyway

bitner (approved):

PR #736, opened 2024-07-19
Feature/add collection search extension V2

vincentsarago:
simpler version of #681 

This PR doesn't add the`/collections - POST` endpoint because:
- it's not mandatory in the spec 
- it will conflict with the transaction extension 😬 

We could in another PR show (in the documentation) how to create a `/collections - POST`  which support both `search and transaction`

cc @tjellicoe-tpzuk

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
 
 ### Added
 * Add Free-text Extension ([#655](https://github.com/stac-utils/stac-fastapi/pull/655))
-* Add Collection-Search Extension
+* Add Collection-Search Extension ([#736](https://github.com/stac-utils/stac-fastapi/pull/736))
 
 ## [3.0.0b2] - 2024-07-09
 
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/request.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/request.py
@@ -4,11 +4,11 @@ from typing import Optional
 
 import attr
 from fastapi import Query
+from stac_pydantic.shared import BBox
 from typing_extensions import Annotated
 
-from stac_pydantic.shared import BBox
 from stac_fastapi.types.rfc3339 import DateTimeType
-from stac_fastapi.types.search import APIRequest, _datetime_converter, _bbox_converter
+from stac_fastapi.types.search import APIRequest, _bbox_converter, _datetime_converter
 
 
 @attr.s
@@ -35,4 +35,3 @@ class FreeTextGetRequest(APIRequest):
         Optional[str],
         Query(description="A basic free-text search parameter."),
     ] = attr.ib(default=None)
-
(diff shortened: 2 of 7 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/request.py
```diff
@@ -0,0 +1,27 @@
...
+    bbox: Optional[BBox] = attr.ib(default=None, converter=_bbox_converter)
+    datetime: Optional[DateTimeType] = attr.ib(
+        default=None, converter=_datetime_converter
+    )
+    limit: Annotated[
+        Optional[int],
+        Query(
+            description="Limits the number of results that are included in each page of the response."  # noqa: E501
+        ),
+    ] = attr.ib(default=None)
```
vincentsarago, line 27: should we put a default to the limit ?

m-mohr: The intention is to behave consistent with Item Search. What does Item Search put?

vincentsarago: Item Search set `10` as default! https://github.com/stac-utils/stac-fastapi/blob/fe4d0dfb9812ec624b5af8bb10e32621b48e090a/stac_fastapi/types/stac_fastapi/types/search.py#L176-L181

vincentsarago:
this is ready for review

CHANGES.md
```diff
@@ -7,8 +7,8 @@
 * add more openapi metadata in input models [#734](https://github.com/stac-utils/stac-fastapi/pull/734)
 
 ### Added
-
-* Add Free-text Extension to third party extensions ([#655](https://github.com/stac-utils/stac-fastapi/pull/655))
+* Add Free-text Extension ([#655](https://github.com/stac-utils/stac-fastapi/pull/655))
+* Add Collection-Search Extension
```
jonhealy1, line 11: Add a link to the PR maybe

jonhealy1 (changes requested):
Cool beans!

CHANGES.md
vincentsarago, line 11:
```suggestion
* Add Collection-Search Extension ([#736](https://github.com/stac-utils/stac-fastapi/pull/736))
```

jonhealy1 (approved):

m-mohr:
@jonhealy1 @vincentsarago I had another look at this and I'm wondering where the queryables enpoint is exposed. The filter extension, which is listed in the conformance classes, would require to be able to query based on queryables but I don't see where the link is added. Is that done in the backends?

vincentsarago:
Yes that should be done by the users in their application.

m-mohr:
@vincentsarago Why though? If you have the filter extension, you need the queryables link anyway. Could return an empty schema, but why this manual step?

Related PR: https://github.com/stac-utils/stac-fastapi-pgstac/pull/89

m-mohr:
@vincentsarago I'm missing the queryables **link** (not the endpoint itself) in `GET /collections`, similar to how it's missing for global Item Search in `GET /` (see https://github.com/stac-utils/stac-fastapi-pgstac/pull/89 ). Maybe I'm missing something though?

vincentsarago:
👍 the links are added in the client as you did in the stac-FastAPI-pgstac PR

PR #739, opened 2024-07-24
add POST - /collections collection-search

vincentsarago:
in addition to the `collection-search` extension work done in #736, this PR adds a `POST - /collections` endpoint in another extension `CollectionSearchPostExtension`

### DIFF
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/collection_search.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/collection_search.py
@@ -1,14 +1,20 @@
 """Collection-Search extension."""
 
 from enum import Enum
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import attr
-from fastapi import FastAPI
+from fastapi import APIRouter, FastAPI
+from stac_pydantic.api.collections import Collections
+from stac_pydantic.shared import MimeTypes
 
+from stac_fastapi.api.models import GeoJSONResponse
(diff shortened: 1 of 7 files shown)


stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/collection_search.py
```diff
@@ -46,7 +52,7 @@ class CollectionSearchExtension(ApiExtension):
             the extension
     """
 
-    GET = CollectionSearchExtensionGetRequest
+    GET: BaseCollectionSearchGetRequest = attr.ib(default=BaseCollectionSearchGetRequest)
```
vincentsarago, line 55: ⚠️ this makes the `GET` attribute *editable* on user input. 

This attribute isn't really used externally or internally because the model is passed to the application using `collections_get_request_model` option, but it make sense

m-mohr:
How does this work in combination with Collection Transactions?

vincentsarago:
@m-mohr it shouldn't be used together, which is why I've created 2 different extension. 

It will be really weird to have a `POST - /collections` being able to handle both the transaction and the search IMO. 

Maybe we can investigate how to do it in another iteration but for now I believe it's fine having both extensions adding their own `POST - /collections` endpoint.

In term of code, it will be really complex (nor impossible) to make them work together because they are both extensions so we can't make them depends on each others

m-mohr:
Ah, I didn't realize that. Thanks for the clarification, sounds good to me.

bitner (approved):

PR #743, opened 2024-07-26
3.0 release

### LINKED ISSUE #678: pin sub-modules (api/extensions) to specific version
The three stac-fastapi sub-modules are linked together, mostly because of the `mono-repo` but also because they depends on each other

- `api` depends on `types` (and not officially on extensions, 😬 circular dependency) 
- `extensions` depends on `api` and `types`

Right now we don't set any version limits of packages and technically this could mean that users could have different version of each packages... and this will not work (it could, but there is a hight chance of breaking changes between the packages).

in https://github.com/stac-utils/stac-fastapi/pull/625#discussion_r1550413449 I've proposed that we pin the sub-module to specific version but we reverted to acquire more opinions (https://github.com/stac-utils/stac-fastapi/pull/625#discussion_r1579318700)

FYI: we do this in `titiler` https://github.com/developmentseed/titiler/blob/main/src/titiler/mosaic/pyproject.toml#L34 and we use `bump2version` to update all the version when we do a release https://github.com/developmentseed/titiler/blob/main/.bumpversion.cfg (which we will introduce in #648)

vincentsarago:
ref #653 

This PR prepare for the release of 3.0
- remove deprecated methods/extensions/functions
- update versions
- pin sub-modules to `~=3.0` TBD (closes #678)


jonhealy1 (changes requested):
Looks good. Will we have to update the changelog?

vincentsarago:
thanks @jonhealy1 

re-requested your review with the changelog updated 

I'm off today so feel free to merge/release this 3.0 version 🥳

jonhealy1:
@vincentsarago have a great vacation!

jonhealy1:
Should we purge the pre-releases from this changelog? All of the changes that took place between 2.5.5.post1 and 3.0.0 would be moved to the 3.0.0 section. This could make it easier to highlight the changes that took place between the major releases?

vincentsarago:
> Should we purge the pre-releases from this changelog? All of the changes that took place between 2.5.5.post1 and 3.0.0 would be moved to the 3.0.0 section. This could make it easier to highlight the changes that took place between the major releases?

we could replicate them or maybe copy/paste them in the migration guide and then add a reference to the guide in the changelog 🤷

vincentsarago:
this is ready for review @jonhealy1 @lossyrob @mmcfarland 🙏

jonhealy1 (approved):

PR #744, opened 2024-08-21
Patch endpoints

rhysrevans3:
**Description:**
Adds `PATCH` endpoints to transaction extension. Adds support for [RFC 6902](https://tools.ietf.org/html/rfc6902) and [RFC 7396](https://tools.ietf.org/html/rfc7396). Pivots on header `Content-Type` value.

Related pull request: https://github.com/stac-api-extensions/transaction/pull/14

**PR Checklist:**


rhysrevans3:
I've had to switch from stac-pydantic `Item`/`Collection` to `Dict` to allow for partial updates. Not sure if this is the best method or if you can switch off pydantic validation. Another option would be having separate models for partial items/collections in stac-pydantic where all attributes.

vincentsarago:
> I've had to switch from stac-pydantic Item/Collection to Dict to allow for partial updates. Not sure if this is the best method or if you can switch off pydantic validation. Another option would be having separate models for partial items/collections in stac-pydantic where all attributes.

Would a typedDict make more sense?

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -0,0 +1,125 @@
...
+class PatchMoveCopy(BaseModel):
+    """Move or Copy Operation."""
+
+    path: str = attr.ib()
+    op: Literal["move", "copy"] = attr.ib()
+
+    def __attrs_init__(self, *args, **kwargs):
+        """Init function to add 'from' field."""
+        super().__init__(*args, **kwargs)
+        self.__setattr__("from", kwargs["from"])
```
vincentsarago, line 88: why `from` is not in the model itself?

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -0,0 +1,125 @@
...
+from stac_pydantic.shared import BBox
+from typing_extensions import Annotated
+
+from stac_fastapi.api.models import CollectionUri, ItemUri
+
+if sys.version_info < (3, 12, 0):
+    from typing_extensions import TypedDict
+else:
+    from typing import TypedDict
+
```
vincentsarago, line 19: I'm pretty sure we can simply do `from typing_extensions import TypedDict` 

typing extensions will use the native TypedDict is available

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -81,6 +87,95 @@ def update_item(
...
+
+        Args:
+            item_id: id of the item.
+            collection_id: id of the collection.
+            patch: either the partial item or list of patch operations.
+
+        Returns:
+            The patched item.
+        """
+        content_type = kwargs["request"].headers.get("content-type", "application/json")
```
vincentsarago, line 109: should we add the `Header` parameter to the input model directly?

vincentsarago: then we could maybe `advertise` which content-type are available?

rhysrevans3: Done 👍

stac_fastapi/types/stac_fastapi/types/transaction.py
rhysrevans3: I followed the import in [types/stac.py](https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/stac.py#L8-L15). Should I update both to `from typing_extensions import TypedDict`?

stac_fastapi/types/stac_fastapi/types/transaction.py
rhysrevans3: I did this as `from` is one of Python's reserved words. There might be a better/neater way to do this.

stac_fastapi/types/stac_fastapi/types/transaction.py
vincentsarago: yes please 🙏

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -385,7 +382,7 @@ async def patch_item(
         Returns:
             The patched item.
         """
-        content_type = kwargs["request"].headers.get("content-type", "application/json")
+        print("content_type", content_type)
```
vincentsarago, line 385: 👀

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -371,6 +367,7 @@ async def patch_item(
         collection_id: str,
         item_id: str,
         patch: Union[PartialItem, List[PatchOperation]],
+        content_type: Optional[str],
```
vincentsarago, line 370: I believe this should be `content_type: Optional[str] = "application/json"

because earlier you had `content_type = kwargs["request"].headers.get("content-type", "application/json")`

vincentsarago: or you can add a line 

```python
content_type = content_type or "application/json"
```

vincentsarago: maybe we could have literal 
```
Literal["application/json-patch+json", "application/merge-patch+json", "application/json"]
```

just to `advertise` what are the supported content type

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -106,6 +106,10 @@ class PatchItem(ItemUri):
         Union[PartialItem, List[PatchOperation]],
         Body(),
     ] = attr.ib(default=None)
+    content_type: Annotated[
+        Optional[str],
```
vincentsarago, line 110: ? `Optional[Literal["application/json-patch+json", "application/merge-patch+json", "application/json"]]`

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -0,0 +1,133 @@
...
+
+@attr.s
+class PostItem(CollectionUri):
+    """Create Item."""
+
+    item: Annotated[Union[Item, ItemCollection], Body()] = attr.ib(default=None)
+
+
+@attr.s
+class PatchAddReplaceTest(BaseModel):
```
vincentsarago, line 62: to me it feels weird to use pydantic model + attrs

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -0,0 +1,133 @@
+"""Transaction extension types."""
```
vincentsarago, line 1: can we move back everything in `stac_fastapi.extensions`? 

by importing `CollectionUri, ItemUri` from `stac_fastapi.api` you create a circular dependency between `types` and `api`. `api` depends on `types` and `types` shouldn't depend on `api`

stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py
```diff
@@ -3,38 +3,23 @@
...
-    """Update Item."""
-
-    item: Annotated[Item, Body()] = attr.ib(default=None)
-
-
-@attr.s
-class PutCollection(CollectionUri):
-    """Update Collection."""
-
-    collection: Annotated[Collection, Body()] = attr.ib(default=None)
```
vincentsarago, line 37: let's try to keep those here

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -0,0 +1,133 @@
...
+    """Patch Collection."""
+
+    patch: Annotated[
+        Union[PartialCollection, List[PatchOperation]],
+        Body(),
+    ] = attr.ib(default=None)
+    content_type: Annotated[
+        Optional[str],
+        Header(),
+    ] = attr.ib(default="application/json")
```
vincentsarago, line 133: PatchCollection and PatchItem should go to `.extensions.transaction`

stac_fastapi/types/stac_fastapi/types/transaction.py
```diff
@@ -0,0 +1,133 @@
...
+    path: str = attr.ib()
+    op: Literal["move", "copy"] = attr.ib()
+
+    def __attrs_init__(self, *args, **kwargs):
+        """Init function to add 'from' field."""
+        super().__init__(*args, **kwargs)
+        self.__setattr__("from", kwargs["from"])
+
+
+PatchOperation = Union[PatchAddReplaceTest, PatchMoveCopy, PatchRemove]
```
vincentsarago, line 91: Those `model` can be moved in `stac.py`

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -20,6 +21,8 @@
 from stac_fastapi.types.requests import get_base_url
 from stac_fastapi.types.rfc3339 import DateTimeType
 from stac_fastapi.types.search import BaseSearchPostRequest
+from stac_fastapi.types.stac import (PartialCollection, PartialItem,
```
vincentsarago, line 24: `stac` is imported at the top so in the code you can use `stac.PartialCollection`

vincentsarago:
@rhysrevans3 should we wait for https://github.com/stac-api-extensions/transaction/pull/14 to be merged/published?

stac_fastapi/types/stac_fastapi/types/stac.py
```diff
@@ -83,3 +75,66 @@ class Collections(TypedDict, total=False):
...
+class PatchMoveCopy(BaseModel):
+    """Move or Copy Operation."""
+
+    path: str
+    op: Literal["move", "copy"]
+
+    def __init__(self, *args, **kwargs):
+        """Init function to add 'from' field."""
+        super().__init__(*args, **kwargs)
+        self.__setattr__("from", kwargs["from"])
```
vincentsarago, line 137: because we're using pydantic model I think we can use Field

```python
class PatchMoveCopy(BaseModel):
   """Move or Copy Operation."""

   path: str
   op: Literal["move", "copy"]
   fromurl: str = Field(alias="from")
```

but then the attribute is `.fromurl` not `.from`

vincentsarago:
@rhysrevans3 I'm trying to get a general sense of the `patch` operation. It's not clear to me what are the requirements that an API should implement. By default it seems that a client will have to support a lot of things 

For example, I'm looking at https://docs.ogc.org/DRAFTS/20-002.html#update_clause and I don't see reference to the `move/copy` operations

stac_fastapi/types/stac_fastapi/types/stac.py
rhysrevans3: Ah I hadn't seen `alias` before that looks neater.

I've also added:
```
model_config = ConfigDict(populate_by_name=True)
```
which means you can do this:
```
patch_operation = PatchMoveCopy(**{"path": "hello", "from":"world", "op": "move"})
```

And overrode the `model_dump` method so `by_alias` defaults to `True` to use the alias in the model dump 
```
patch_operation.model_dump()
#> {'path': 'hello', 'op': 'move', 'from': 'world'}
```
Not sure if this is the right thing to do or if we should expect people to use:
```
patch_operation.model_dump(by_alias=True)
```

rhysrevans3:
My understanding is [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789), which the [OGC doc](https://docs.ogc.org/DRAFTS/20-002.html#update_clause) references, say a PATCH endpoint should except a set of actions to modify the requested resource. But it doesn't specify what these actions should be.

[RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902) extends this and specify a set of allowed operations (add, remove, replace, move, copy, test) which is what I have suggested in https://github.com/stac-api-extensions/transaction/pull/14.

Currently the [README](https://github.com/stac-api-extensions/transaction/blob/main/README.md?plain=1#L89) on the transaction extension suggests the PATCH endpoint follows [RFC 7396](https://datatracker.ietf.org/doc/html/rfc7396) which accepts a partial json document which it merges with the existing document this might not be the best choice for STAC because of it's handling of `null` values. As mentioned in this issue https://github.com/stac-api-extensions/transaction/issues/13

vincentsarago:
@rhysrevans3 The overall looks great, just added one comment about using stac-pydantic base model.

As I've asked couple weeks ago, should we wait for https://github.com/stac-api-extensions/transaction/pull/14 to be merged? do you know anyone who could help move this PR forward?

stac_fastapi/types/stac_fastapi/types/stac.py
```diff
@@ -83,3 +75,68 @@ class Collections(TypedDict, total=False):
...
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    path: str
+    op: Literal["move", "copy"]
+    from_: str = Field(alias="from")
+
+    def model_dump(self, by_alias=True, **kwargs) -> Dict[str, Any]:
+        """Override by_alias default to True"""
+        return super().model_dump(by_alias=by_alias, **kwargs)
```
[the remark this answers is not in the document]
rhysrevans3: Ah that's perfect thank you 👍

rhysrevans3:
> @rhysrevans3 The overall looks great, just added one comment about using stac-pydantic base model.
> 
> As I've asked couple weeks ago, should we wait for [stac-api-extensions/transaction#14](https://github.com/stac-api-extensions/transaction/pull/14) to be merged? do you know anyone who could help move this PR forward?

Ideally yes I think it would be better to have that PR merged first.

But my understanding is that [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902) and [RFC 7396](https://datatracker.ietf.org/doc/html/rfc7396) are implementations of [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789), which is the current STAC/OGC specification. So this merging this pull still keeps within the current STAC specification.

But given @m-mohr comments on that pull request maybe it's best to wait for OGC's opinion.

m-mohr:
Did someone open an issue with OGC API - Features? (If yes, where?) Otherwise you may wait forever ;)

rhysrevans3:
Sorry I've been at a conference/on leave I've created a new issue [here](https://github.com/opengeospatial/ogcapi-features/issues/957)

m-mohr:
@rhysrevans3 Thanks.

Given the response in https://github.com/opengeospatial/ogcapi-features/issues/957, we may also want to open an issue in the Transaction extension itself so that we can reflect the updated to OGC API - Features in the extension, too. Would you mind opening that issue, too?

m-mohr:
Oh cool, too many repos to monitor. Left my comment(s) there.

rhysrevans3:
@vincentsarago [The transaction extension pull request](https://github.com/stac-api-extensions/transaction/issues/13) is just waiting for one more review. But I wondered if we could get this one merged in first? I'm confident the changes will be accept and this doesn't break the existing transaction extension.

vincentsarago:
hi @rhysrevans3 

I'm a bit concerned that if merging this PR as is it would introduce a breaking change making `merge_patch_item`, `json_patch_item`, `merge_patch_collection` and `json_patch_collection` mandatory  methods in any Transaction client 

Also reading the spec 

> The default standard for PATCH is [RFC 7396](https://tools.ietf.org/html/rfc7396).
 Optionally, in addition to RFC 7396, [RFC 6902](https://tools.ietf.org/html/rfc6902) can be supported with the two methods
 being distinguished via the Accept/Content-Type header.

This sounds to me that only RFC7396 (merge) should be implemented and RFC6902 (patch) should be optional, but because of the 4 mandatory methods this is not reflected. 


I understand correctly, the current implementation is not following the transaction spec because we don't have `PATCH` support https://github.com/stac-utils/stac-fastapi/blob/20ae9cfaf87ed892ef3235d979892e7e24c63fc0/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py#L151-L154

Note: IMO we should move everything related to `transaction` to the `extension` module (https://github.com/stac-utils/stac-fastapi/blob/88639230e3bd1573e23e3a646204cd85932079cd/stac_fastapi/types/stac_fastapi/types/core.py#L40C1-L621C1). But it might be easier to do it after merging this PR

rhysrevans3:
@vincentsarago you're right I had missed that. I've updated `json_patch_item` and  `json_patch_collection` to raise an `NotImplementedError` as a default. Is that enough or does it need to be in the registration of the endpoint?

Happy to use this pull request to move everything to the `transaction` to the `extension module` if required.

rhysrevans3:
I could also add a `NotImplemented` default to the `merge-patch` to not break existing implementations? I'm unsure if the Transaction extension _requires_ a PATCH endpoint or if you could specify a different `Content-Type`. For example if you wanted to implement `json-patch` and not `merge-patch` or even a non-JSON PATCH like [XML PATCH RFC 5261](https://datatracker.ietf.org/doc/html/rfc5261)

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -80,6 +81,100 @@ def update_item(
         """
         ...
 
+    def patch_item(
```
vincentsarago, line 84: What do you think about making `patch_item` the abstract method. Having `patch_item` depending on two abstract method makes things a bit complex IMO and will make customization a bit difficult.

rhysrevans3: Yes that sounds sensible to me. I'll make the changes now!

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -370,18 +237,13 @@ async def update_item(
...
         item_id: str,
-        patch: Union[stac.PartialItem, List[stac.PatchOperation]],
-        content_type: Optional[
-            Literal[
-                "application/json-patch+json",
-                "application/merge-patch+json",
-                "application/json",
-            ]
-        ] = "application/json",
+        patch: Any,
```
vincentsarago, line 245: could `patch` still be of type `Union[stac.PartialItem, List[stac.PatchOperation]],` ?

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -80,6 +80,29 @@ def update_item(
         """
         ...
 
+    @abc.abstractmethod
+    def patch_item(
+        self,
+        collection_id: str,
+        item_id: str,
+        patch: Any,
+        content_type: Optional[str],
```
vincentsarago, line 89:
```suggestion
        content_type: Optional[str] = None,
```

stac_fastapi/types/stac_fastapi/types/core.py
rhysrevans3: It could but other PATCH implementations like [XML PATCH](https://datatracker.ietf.org/doc/html/rfc5261) would have a different type so for future proofing I set it to `Any`. But can switch back if we think most will use JSON merge or patch. I guess this can be overwritten in the backend if other types are needed?

vincentsarago: 🤔 the spec doesn't mention XML patch but only `RFC 7396` and `RFC 6902`. If we ever add `RFC 5261` to the spec we could update it here but for now I think it's safe to exclude it

stac_fastapi/types/stac_fastapi/types/core.py
vincentsarago: I think `None` is fine. 

we could also totally remove it from the arg list and use `request: Request`

rhysrevans3: That was my attempt to make it configurable in the `api.html` page but it doesn't seem like it actually sets the Content-Type so I'd be happy to go back to using request if that's better for stac-fastapi.

Screenshot 2025-04-03 at 09 27 50

vincentsarago: https://github.com/fastapi/fastapi/discussions/7786#discussioncomment-9227946 

I think you can achieve this with openapi customization 

```python
def register_patch_item(self):
    """Register patch item endpoint (PATCH
    /collections/{collection_id}/items/{item_id})."""
    self.router.add_api_route(
        name="Patch Item",
        path="/collections/{collection_id}/items/{item_id}",
        response_model=Item if self.settings.enable_response_models else None,
        responses={
            200: {
                "content": {
                    MimeTypes.geojson.value: {},
                },
                "model": Item,
            }
        },
        openapi_extra={
            "requestBody": {
                "content": {
                    "application/json-patch+json": {
                        "schema": ...,
                    },
                    "application/merge-patch+json": {
                        "schema": ...,
                    },
                    "application/json": {
                        "schema": ...,
                    }
                },
                "required": True,
            },
        },
        response_class=self.response_class,
        response_model_exclude_unset=True,
        response_model_exclude_none=True,
        methods=["PATCH"],
        endpoint=create_async_endpoint(
            self.client.patch_item,
            PatchItem,
        ),
    )


    

def register_patch_collection(self):
    """Register patch collection endpoint (PATCH /collections/{collection_id})."""
    self.router.add_api_route(
        name="Patch Collection",
        path="/collections/{collection_id}",
        response_model=Collection if self.settings.enable_response_models else None,
        responses={
            200: {
                "content": {
                    MimeTypes.geojson.value: {},
                },
                "model": Collection,
            }
        },
        openapi_extra={
            "requestBody": {
                "content": {
                    "application/json-patch+json": {
                        "schema": ...,
                    },
                    "application/merge-patch+json": {
                        "schema": ...,
                    },
                    "application/json": {
                        "schema": ...,
                    }
                },
                "required": True,
            },
        },
        response_class=self.response_class,
        response_model_exclude_unset=True,
        response_model_exclude_none=True,
        methods=["PATCH"],
        endpoint=create_async_endpoint(
            self.client.patch_collection,
            PatchCollection,
        ),
    )
```

I'll let you figure out the `schemas` 😅 

I think if would be fine to `advertise support for the 3 content-type `application/json-patch+json`, `application/merge-patch+json` and `application/json` and then let the application raise `NotImplementedError` within the client method... or we make this configurable... or we just document how people could add the `merge` content-type

rhysrevans3: Thanks that's just what I was looking! I've added `openapi_extra` for patch item and collection and removed the `content_type` parameter.

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -133,6 +155,26 @@ def update_collection(
         """
         ...
 
+    @abc.abstractmethod
+    def patch_collection(
+        self,
+        collection_id: str,
+        patch: Union[stac.PartialItem, List[stac.PatchOperation]],
```
vincentsarago, line 162: should the type be 

```python
patch: Union[stac.PartialCollection, List[stac.PatchOperation]],
```

stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -246,6 +310,26 @@ async def update_collection(
         """
         ...
 
+    @abc.abstractmethod
+    async def patch_collection(
+        self,
+        collection_id: str,
+        patch: Union[stac.PartialItem, List[stac.PatchOperation]],
```
vincentsarago, line 317: same as https://github.com/stac-utils/stac-fastapi/pull/744/files#r2026723598

rhysrevans3: Good catch! Poor copy/paste job by me sorry

rhysrevans3:
I've written a conversion from `merge` to `patch` for the elasticsearch-opensearch backend if this is useful to other backends? I could move it to the core utils?

https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/blob/2b1381821ec978ea254df0c8fae87420186a8789/stac_fastapi/core/stac_fastapi/core/utilities.py#L145-L171

vincentsarago:
@rhysrevans3 yeah that would be great. It will lower the barrier for people to support both patch and merge 🙏 

Maybe it could be a method on the PartialItem/PartialCollection model 

☝️ while writing this I realized that they are typedDict not pydantic model 🤔

rhysrevans3:
> @rhysrevans3 yeah that would be great. It will lower the barrier for people to support both patch and merge 🙏
> 
> Maybe it could be a method on the PartialItem/PartialCollection model
> 
> ☝️ while writing this I realized that they are typedDict not pydantic model 🤔

Yes I copied the Item/Collection models in the same file. I think the reason for using TypeDict over Pydantic models was performance (avoid validation) I remember there was some discussion when we moved to pydantic v2 #625 which suggested that's no longer an issue? Should I switch to using Pydantic models?

vincentsarago:
>  Should I switch to using Pydantic models?

Yes we should use pydantic model for input validation. It's only for the `output` that we have optional validation

rhysrevans3:
> @rhysrevans3 I'm sorry it's taking longer 😅
> 
> I now think (with the addition of a `merge to patch` method) that by default we should make both patch/merge supported by default 😅.
> 
> On think I'm still not sure is how the merging/patching is made.
> 
> I'm looking at https://github.com/rhysrevans3/stac-fastapi-elasticsearch/blob/2b1381821ec978ea254df0c8fae87420186a8789/stac_fastapi/elasticsearch/stac_fastapi/elasticsearch/database_logic.py#L940-L1031 to try to understand what's going on.
> 
> We should write some pseudo code in the doc (in the docstring first) to show how the `patch` method will look like
> 
> ```python
>     def patch_item(
>         self,
>         collection_id: str,
>         item_id: str,
>         patch: Union[stac.PartialItem, List[stac.PatchOperation]],
>         **kwargs,
>     ) -> Optional[Union[stac.Item, Response]]:
>         """Update an item from a collection.
> 
>         Called with `PATCH /collections/{collection_id}/items/{item_id}`
> 
>         Args:
>             item_id: id of the item.
>             collection_id: id of the collection.
>             patch: either the partial item or list of patch operations.
> 
>         Returns:
>             The patched item.
>         """
>         # convert patch item to list of merge operations
>         if isinstance(patch, PartialItem):
>             patch = merge_to_operations(patch)
> 
>         # Get Original Item
>         item = backend.get_item(collection_id, item_id)
> 
>         # Update Item body from merge op
>         item = update_item(item, patch)
> 
>         # Push Item to the backend
>         _ = backend.create_item(collection_id, item_id, item)
> 
>         return item
> ```

No worries I'd rather it takes a bit longer and we get it right!

For Elasticsearch you can update records using Java style scripts so for each operation type I've written an equivalent Elasticsearch script. I'm not sure if you could do the same for PostgreSQL or the other backends.

A merge is equivalent to a group of add and remove operations (null values are used to remove). so you can convert from merge to operations to Elasticsearch.

stac_fastapi/types/stac_fastapi/types/stac.py
```diff
@@ -188,3 +153,78 @@ class PatchMoveCopy(StacBaseModel):
...
+        Returns:
+            List[PatchOperation]: Equivalent list of RF6902 operations
+        """
+        return self.merge_to_operations(self.model_dump())
+
+
+class PartialCollection(BasePartial):
+    """Partial STAC Collection."""
+
+    type: Optional[str]
```
vincentsarago, line 201: not sure why/how the test passes but for Optional we need to set `= None` with pydantic 2.0

stac_fastapi/extensions/tests/test_transaction.py
```diff
@@ -46,6 +47,19 @@ def update_item(self, collection_id: str, item_id: str, item: Item, **kwargs):
...
         }
 
+    def patch_item(
+        self,
+        collection_id: str,
+        item_id: str,
+        patch: Union[PartialItem, List[PatchOperation]],
+        **kwargs,
+    ):
+        return {
```
vincentsarago, line 57: I think we should also test `PartialItem.merge_to_operations` here

rhysrevans3: The conversion was missing of the initial `/` on the path so these tests were a good addition thanks.

rhysrevans3:
I've not tested this but this is an example of running a list patch operations on an item:

```python
from stac_fastapi.types.stac import PatchOperation
from stac_pydantic import Item

def patch_item(item: Item, operations: PatchOperation) -> Item:
    item_dict = item.model_dump()

    for operation in operations:
        path_parts = operation.path.split('/')
 
       if operation.op == "test":
            test_value = item_dict.copy()
            for path_part in path_parts:
                test_value = test_value[path_part]

            assert test_value == operation.value
            continue

        if operation.op == "replace":
            nest = item_dict.copy()
            for path_part in path_parts:
                assert path_part in nest
                nest = nest[path_part]

        update = {}

        if operation.op in ["add", "copy", "replace", "move"]:
            if operation.hasattr("from_"):
                from_parts = operation.from_.split('/')

                value = item_dict.copy()
                for path_part in from_parts:
                    value = value[path_part]

            else:
                value = item.value

            update = value
            for path_part in path_parts.reverse():
                update = {path_part: update}

        if operation.op in ["remove", "move"]:
            if operation.op == "move":
                path_parts = from_parts
 
            last_part = path_parts.pop(-1)

            nest = item_dict
            for path_part in path_parts:
                nest = nest[path_part]

            del nest[last_part]

        return Item.model_validate(item_dict | update)
```

vincentsarago:
FYI: sorry for letting this PR stale. Because it adds breaking changes I'll wait for the next breaking change release to be planned before merging.

rhysrevans3:
@vincentsarago thanks for staying on top of this. Including it in the next major release sounds sensible 😃

stac_fastapi/types/stac_fastapi/types/stac.py
```diff
@@ -77,3 +79,156 @@ class Collections(TypedDict):
...
+    description: Optional[str] = None
+    links: Optional[Dict[str, Any]] = None
+    keywords: Optional[List[str]] = None
+    license: Optional[str] = None
+    providers: Optional[List[Dict[str, Any]]] = None
+    extent: Optional[Dict[str, Any]] = None
+    summaries: Optional[Dict[str, Any]] = None
+    assets: Optional[Dict[str, Any]] = None
+    numberMatched: Optional[int] = None
+    numberReturned: Optional[int] = None
```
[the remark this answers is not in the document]
rhysrevans3: Sorry not sure how these made it there. I've removed them now.

PR #746, opened 2024-08-27
revert to fastapi and update version

vincentsarago:
FastAPI reverted their changed and now doesn't include *useless* dependencies https://github.com/fastapi/fastapi/discussions/11525 

we can move back to `fastapi` instead of `fastapi-slim` and also relaxe the version dependency!


jonhealy1 (approved):
Looks great

stac_fastapi/types/setup.py
```diff
@@ -6,7 +6,7 @@
     desc = f.read()
 
 install_requires = [
-    "fastapi-slim>=0.111.0",
+    "fastapi>=0.100.0",
```
vincentsarago, line 9:
```suggestion
    "fastapi>=0.109.0",
```

because we use new features introduced in starlette 0.29

CHANGES.md
```diff
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+## [3.0.1] - 2024-08-27
+
+### Changed
+
+* Update fastapi version to `>=0.100.0`
```
vincentsarago, line 9:
```suggestion
* Update fastapi version to `>=0.109.0`
```

Issue #761, opened 2024-10-14
Add STAC_API_VERSION in landing page


m-mohr:
I'm a bit confused by https://github.com/stac-utils/stac-fastapi/blob/3c4ce2e0b8616d388b94a9cfb2a28e5d44ada29a/stac_fastapi/types/stac_fastapi/types/core.py#L271

STAC_API_VERSION is meant to be the STAC API version (i.e. stac-api-spec), right?

But the STAC API Version is different to the STAC version. We try to expose STAC 1.1.0 items and update the landing page to 1.1.0, too, but setting the STAC_API_VERSION seems incorrect, because there is no 1.1.0 version for the API.

m-mohr:
Maybe it's also valuable to have STAC_API_VERSION if you want to switch between (bugfix) API versions, but having a separate STAC_VERSION is certainly a good idea. Do you think it makes sense to also have a STAC_API_VERSION? (e.g. to update to easily/quickly STAC API 1.0.1 once released?)

vincentsarago:
the stac-api-specification only mention `stat-version` in the landing model 

https://github.com/radiantearth/stac-api-spec/blob/bab9bdf22b54cd119e931859b0da13f42091abb8/core/commons.yaml#L50-L51

 but extra values are allowed https://github.com/stac-utils/stac-pydantic/blob/main/stac_pydantic/catalog.py#L22

vincentsarago:
I've created #762

adding `stac_api_version` is a bit more complication because it will involve more changes and a new Minor version so if we really want this we should create a specific issue for it

m-mohr:
Not feeling strongly about STAC_API_VERSION, we can just fix the actual issue for now. Thanks for creating the PR!

vincentsarago:
closing this for now because it's not a spec requirement

PR #762, opened 2024-10-14
update stac-version

vincentsarago:
ref #761

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,10 @@
 * Add `numberMatched` and `numberReturned` properties in `types.stac.ItemCollection` model
 * Add `numberMatched` and `numberReturned` properties in `types.stac.Collections` model
 
+## Changed
+
+* use `stac_pydantic.version.STAC_VERSION` instead of `stac_pydantic.api.version.STAC_API_VERSION` as application `stac_version`
+
 ## [3.0.3] - 2024-10-09
 
 ### Removed
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -10,8 +10,8 @@ from fastapi.openapi.utils import get_openapi
 from fastapi.params import Depends
 from stac_pydantic import api
 from stac_pydantic.api.collections import Collections
-from stac_pydantic.api.version import STAC_API_VERSION
 from stac_pydantic.shared import MimeTypes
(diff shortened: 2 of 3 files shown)


vincentsarago:
@m-mohr I don't think there is any env variable that can change what is set in https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/api/stac_fastapi/api/app.py#L96 
https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/core.py#L287

Worth noting that I don't think this should be configurable because it all comes down to what version for the spec is supported by `stac-pydantic` 

I still think that it might be good to indicate which stac and stac-api version the application provide (from stac-pydantic)

m-mohr:
Oh dang, I thought that's a env variable similar to STAC_API_TITLE etc. 
Seems like a misunderstanding then. In 1.0 and 1.1 the changes are not that big so I guess just updating the version number to 1.1.0 for some instances could make sense...

gadomski (approved):

vincentsarago:
> So +1 to: "I still think that it might be good to indicate which stac and stac-api version the application provide (from stac-pydantic)" - could that be simply an env variable?

So there are 2 things:
- STAC VERSION: should indicate the STAC version specification returned by the endpoints, those should match what stac-pydantic has because we use it (optional) for validation, so I can't think about a situation where the STAC Version number would be different from the one from stac-pydantic 
- STAC API VERSION: this is almost the same (models used to validate response) but add also the spec for what endpoints are available. I'm having hard time to think why/how user would have a stac-fastapi based application which don't follow the version forwarded from stac-pydantic 

> could that be simply an env variable?

short answer: I don't think so

jonhealy1:
A stac api should work with any stac version, even a catalog that may be a mix of v1 or v1.1?

m-mohr:
Right, so we need stac-pydantic to support STAC v1.1.0 soon.

vincentsarago:
@jonhealy1 

> A stac api should work with any stac version, even a catalog that may be a mix of v1 and v1.1?

🤷 depends if the 1.1 is backward compatible with 1.0

m-mohr:
It is, I'd say... otherwise it should've been a 2.0.

vincentsarago:
IMO: if someone wants to change the STAC VERSION, it means they have a custom version of stac-fastapi, which also means they don't really need to control this with an Env variable

PR #763, opened 2024-10-15
add stac-api-version in landing page response

### LINKED ISSUE #761: Add STAC_API_VERSION in landing page

vincentsarago:
closes #761 

This PR adds a `stac_api_version` attribute in the `stac.LandingPage` model, returned by `/` endpoint.

**This is not in the specification**, in fact it's not clear to me what is in the specification for the landing page response! 

```json
{
  "type": "Catalog",
  "id": "stac-fastapi",
  "title": "stac-fastapi",
  "description": "stac-fastapi",
  "stac_version": "1.0.0",
  "stac_api_version": "1.0.0",
  "conformsTo": [
    ...
  ],
  "links": [
    ...
  ],
  "stac_extensions": []
}
```

The pydantic model from stac-pydantic, allows extra attribute so this addition should not `break` any thing, and I don't think we need to add this in the model as well. It will be specific to stac-fastapi applications

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 
 * Add `numberMatched` and `numberReturned` properties in `types.stac.ItemCollection` model
 * Add `numberMatched` and `numberReturned` properties in `types.stac.Collections` model
+* add `stac_api_version` in Landing page response
 
 ## Changed
 
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -10,6 +10,7 @@ from fastapi.openapi.utils import get_openapi
 from fastapi.params import Depends
 from stac_pydantic import api
 from stac_pydantic.api.collections import Collections
+from stac_pydantic.api.version import STAC_API_VERSION
 from stac_pydantic.shared import MimeTypes
 from stac_pydantic.version import STAC_VERSION
 from starlette.middleware import Middleware
@@ -94,6 +95,7 @@ class StacApi:
         )
     )
     stac_version: str = attr.ib(default=STAC_VERSION)
+    stac_api_version: str = attr.ib(default=STAC_API_VERSION)
     description: str = attr.ib(
         default=attr.Factory(
             lambda self: self.settings.stac_fastapi_description, takes_self=True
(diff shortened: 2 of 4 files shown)


jonhealy1:
Should there be a field(s) that can be set to say like 'stac-fastapi-pgstac' and version '3.0.0' or something to specify the backend type/version?

m-mohr:
This is a misunderstanding. You can parse the STAC API Version from the conformance classes, so we don't need a separate property. I was more thinking that it might be useful to make stac-fastapi more configurable towards supporting multiple versions and that users may want to choose which they want to host.

vincentsarago:
> Should there be a field(s) that can be set to say like 'stac-fastapi-pgstac' and version '3.0.0' or something to specify the backend type/version?

@jonhealy1, there is already https://github.com/stac-utils/stac-fastapi/blob/8235d919b3ab742838d628653f14a54ae694cfe6/stac_fastapi/types/stac_fastapi/types/config.py#L22-L25

which is then used in openAPI https://github.com/stac-utils/stac-fastapi/blob/8235d919b3ab742838d628653f14a54ae694cfe6/stac_fastapi/api/stac_fastapi/api/app.py#L395-L401

vincentsarago:
> This is a misunderstanding, I never intended to have a separate property. You can parse the STAC API Version from the conformance classes, so we don't need a separate property. I was more thinking that it might be useful to make stac-fastapi more configurable towards supporting multiple versions and that users may want to choose which they want to host.

@m-mohr ah! 

well `stac_version` should be set anyway to reflect the stac spec version so #762 was 👍 

>  I was more thinking that it might be useful to make stac-fastapi more configurable towards supporting multiple versions and that users may want to choose which they want to host.

I'm not even sure what supporting multiple versions means to you. Items retrieved from backend can be of multiple version, but stac-fastapi application is designed with models (stac-pydantic) which are for a specific version, which gets advertised with `stac_version` in the landing page

vincentsarago:
I feel we don't really need this then!

If the stac-api change and add this property we will add it

Issue #645, opened 2024-03-29
Document new enhanced middleware configuration options


jonhealy1:
Description:

The recent enhancements to middleware configuration introduced in https://github.com/stac-utils/stac-fastapi/pull/442 add significant flexibility and functionality to the StacApi class, allowing for more granular control over middleware behavior and dynamic addition after application initialization. These changes are a valuable addition to our project and should be clearly documented for users to understand and leverage effectively.

The PR contains a detailed description and examples that could serve as a basis for updating our README. Incorporating this information into the README will ensure our documentation stays current with the codebase and provides clear guidance to users on utilizing the new middleware configuration options.

Action Items:

- Summarize the enhancements and their benefits in a user-friendly manner.
- Include examples of how to configure and dynamically add middleware using the new approach.
- Highlight any potential breaking changes and recommend steps for migration.

vincentsarago:
https://stac-utils.github.io/stac-fastapi/migrations/v3.0.0/#middlewares-configuration

PR #764, opened 2024-10-17
update CORS documentation

vincentsarago:
ref https://github.com/stac-utils/stac-fastapi/issues/645


m-mohr:
Thank you @vincentsarago

Issue #765, opened 2024-12-02
Allow multiple security dependency per route 


rhysrevans3:
@pedro-cf raised an issue on the stac-fastapi-elasticsearch-opensearch https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/issues/310 that you can't use multiple security route dependency. 

Use case for this would be if someone want to use multiple OAuth flows at the same time.

This is a known issue with FastAPI the suggested solution is to use another dependency to "merge" the different dependencies.

vincentsarago:
@rhysrevans3 I'm not sure to fully get what the issue with `stac-fastapi` and how to solve it 🤷

rhysrevans3:
@vincentsarago I made this issue on `stac-fastapi` as this is where the dependency are injected and I thought this might be a common issue for all of the backends.

I'll close this here and more to the elasticsearch repo thanks for the feedback.

rhysrevans3:
@alukach thanks for the feedback. Yes sorry "issue" was probably the wrong terminology and merging dependencies was what I meant by 

> the suggested solution is to use another dependency to "merge" the different dependencies.

I did have a pull request https://github.com/stac-utils/stac-fastapi/pull/766 where I suggested separating the auth route dependencies and automatically merging them if there are multiple for a single route. But I'm in the process of moving this to the `stac-fastapi-elasticsearch-opensearch` repo.

The main reason for merging dynamically was to allow the `*` default route to still be used with multiple auth dependencies.

For example I want to use Basic Auth for my admin user to be able to access all endpoints but my reader client, that uses a different OAuth flow, to only access the search endpoint.

Without the default route I would have to write a route for every endpoint and write my own merge dependency for search to include both flows. Which works but I would say, it makes the auth configuration more difficult to work with. And could increase the chance of human errors, like missing an endpoint.

Maybe there is a way to merge them and still use the default route through more complex authorisation.

PR #766, opened 2024-12-02
Separating route securities from other dependencies. 

rhysrevans3:
**Related Issue(s):**
- #765

**Description:**
Currently if you have more than one security dependency on a single route, if either fail then the route will be unauthorised. This is a known issue in FastAPI and the suggested solution is to use a single dependency to merge the sub dependencies. This makes configuration more difficult and removes the ability to use default security dependencies.

Suggested solution is to separate route securities from other dependencies. This then allows them to be merged when dynamically.

There are two options for merging in this pull request.

`merge_dependencies1` uses the traditional merging method but this requires dependencies return a value or `None` rather than raise an error. This means it would need to be used for all routes even those without multiple dependencies.

`merge_dependencies2` calls each dependency itself which allows them to raise their own errors so can be used only on routes that have multiple dependencies.

**PR Checklist:**

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -1,6 +1,5 @@
 """Fastapi app creation."""
 
-
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import attr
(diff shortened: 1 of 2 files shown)


rhysrevans3:
Moving to `stac-fastapi-elasticsearch-opensearch` see issue #765 for details.

Issue #750, opened 2024-09-12
Unclear error message for failed response model validation


drnextgis:
When response model validation is enabled, the API returns an unclear error message that doesn't specify which property failed validation. For example:
```
{"code":"AttributeError","description":"'int' object has no attribute 'startswith'"}
```
In this case, the issue was caused by the asset href containing a number instead of a string.

vincentsarago:
@drnextgis can you be a bit more verbose. which version of stac-fastapi, which backend, which endpoint?

vincentsarago:
Oh I see, you're saying we should try to make the error message a bit more verbose 😓 

I guess https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/api/stac_fastapi/api/errors.py#L49-L97 needs to be updated

j08lue:
Looks like there is a [`ValidationError.json`](https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.ValidationError.json) 👌

PR #772, opened 2025-01-07
return more information when pydantic raise validation error

vincentsarago:
ref https://github.com/stac-utils/stac-fastapi/issues/750

```
>> httpx -m POST http://127.0.0.1:8000/search -j '{"limit": "yo"}' 
{
    "detail": [
        {
            "type": "int_parsing",
            "loc": [
                "body",
                "limit"
            ],
            "msg": "Input should be a valid integer, unable to parse string as an integer",
            "input": "yo"
        }
    ],
    "body": {
        "limit": "yo"
    }
}

>> httpx -m GET http://127.0.0.1:8000/search\?limit\=yo
{
    "detail": [
        {
            "type": "int_parsing",
            "loc": [
                "query",
                "limit"
            ],
            "msg": "Input should be a valid integer, unable to parse string as an integer",
            "input": "yo"
        }
    ],
    "body": null
}
```

@drnextgis is that better ?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 ## Changed
 
 * use `stac_pydantic.version.STAC_VERSION` instead of `stac_pydantic.api.version.STAC_API_VERSION` as application `stac_version`
+* Return more informations from pydantic validation errors
 
 ## [3.0.3] - 2024-10-09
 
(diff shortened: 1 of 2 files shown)


drnextgis:
That is definitely much better!

Issue #753, opened 2024-09-23
remove `post_request_model` attribute in the BaseCoreClient and AsyncBaseCoreClient


vincentsarago:
https://github.com/stac-utils/stac-fastapi/blob/753341f18d6c3d0454efef52d0ff841cfa3215f7/stac_fastapi/types/stac_fastapi/types/core.py#L344

This attribute is not used in the abstract methods and is not event guarantee to be used in the client. We should let backend define what attribute they need 

cc @hrodmn

PR #773, opened 2025-01-07
deprecate post_request_model in `BaseCoreClient` and `AsyncBaseCoreClient`

vincentsarago:
ref https://github.com/stac-utils/stac-fastapi/issues/753

The attribute is not use in any abstract method. It's up to the backend Client to add desired attributes

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/core.py
+++ b/stac_fastapi/types/stac_fastapi/types/core.py
@@ -1,6 +1,7 @@
 """Base clients."""
 
 import abc
+import warnings
 from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urljoin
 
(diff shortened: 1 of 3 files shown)


stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -341,7 +342,20 @@ class BaseCoreClient(LandingPageMixin, abc.ABC):
...
+        """Check and raise warning if `post_request_model` is set."""
+        if value is not None:
+            warnings.warn(
+                "`post_request_model` attribute is deprecated and will be removed in 3.1",
+                DeprecationWarning,
+            )
+
+    def __attrs_post_init__(self):
+        """Set default value for post_request_model."""
+        self.post_request_model = self.post_request_model or BaseSearchPostRequest
```
vincentsarago, line 358: this might be a breaking change after all 😬 

I need to be able to set this as default but also make sure we can raise a warning if users are setting it 

🤔

Issue #778, opened 2024-12-19
`/search`ing by date and/or datetime and/or datetime+ms


jgaucher-cs:
Hi, when calling the /search endpoint that is implemented here: https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/api/stac_fastapi/api/app.py, it seems we cannot search by date, but only by datetime. 

A GET request with `/search?datetime=2000-01-01` will fail with this error from https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/rfc3339.py
```
b'{"detail":"Invalid RFC3339 datetime."}'
```
A POST request will fail with what seems to by a pydantic error: 
```
b'{"detail":[{"type":"datetime_parsing","loc":["body","datetime"],"msg":"Input should be a valid datetime, invalid datetime separator, expected `T`, `t`, `_` or space","input":"2000-01-01","ctx":{"error":"invalid datetime separator, expected `T`, `t`, `_` or space"}}]}'
```
We would like to allow our users to search by date without time, then our application (that uses stac_fastapi) would return all products for this whole day. Would this be possible ?

Then we would like to do the same for milliseconds: our products have a ms information, so if the user searches by datetime+ms, we would like to return products only for this specific ms. If he searches by datetime without ms, we would like to return all products for this whole second.

Our problem is that, if he searches for ms=.000 (to have products only for this specific ms) with a GET request e.g. `/search?datetime=2000-01-01T01:01:01.000Z`, stac_fastpi **removes the ms=.000 information**: our app receives the value `datetime=2000-01-01T01:01:01Z`, so it doesn't know anymore if the user wanted only products for this specific ms, or for this whole second. I think this removal is made when stac_fastapi converts the str to datetime in `rfc3339_str_to_datetime` https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/types/stac_fastapi/types/rfc3339.py#L51 (I couldn't find where this datetime is converted back to str).

To summarize, my two questions are:

1. Could a user search by date, not datetime ?
2. Could we stop stac_fastpi from removing the "ms=.000" information ?

Thank you :)

vincentsarago:
🤦 sorry for not following @jgaucher-cs 

I get it now. I think this could be considered as a BUG, especially if POST/GET request do not behave the same. 

I'm going to open a PR to propose some change (**breaking change**)

PR #780, opened 2025-01-14
keep datetime as string type in GET Request models

vincentsarago:
**Related Issue(s):**

- https://github.com/stac-utils/stac-fastapi/issues/778

**Description:**

This PR change the `type` for the datetime parameter in GET request for `/search`, `/collection` and `/items` endpoints.

with this change, The GET and POST request models behave similarly by setting type for `string`.

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/models.py
+++ b/stac_fastapi/api/stac_fastapi/api/models.py
@@ -1,5 +1,6 @@
 """Api request/response models."""
 
+from datetime import datetime as dt
 from typing import List, Optional, Type, Union
 
 import attr
@@ -9,14 +10,13 @@ from stac_pydantic.shared import BBox
 from typing_extensions import Annotated
 
 from stac_fastapi.types.extension import ApiExtension
-from stac_fastapi.types.rfc3339 import DateTimeType
+from stac_fastapi.types.rfc3339 import str_to_interval
 from stac_fastapi.types.search import (
     APIRequest,
     BaseSearchGetRequest,
     BaseSearchPostRequest,
     Limit,
     _bbox_converter,
-    _datetime_converter,
 )
 
 try:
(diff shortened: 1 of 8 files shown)


stac_fastapi/api/stac_fastapi/api/models.py
```diff
@@ -121,9 +121,40 @@ class ItemCollectionUri(APIRequest):
...
+        interval = str_to_interval(self.datetime)
+        return interval if isinstance(interval, dt) else interval[0]
+
+    @property
+    def end_date(self) -> Optional[dt]:
+        """End Date."""
+        if self.datetime is None:
+            return self.datetime
+        interval = str_to_interval(self.datetime)
+        return interval[1] if isinstance(interval, tuple) else None
```
gadomski, line 157: Feels funny that these conditionals are inverted:

```suggestion
        return interval[0] if isinstance(interval, tuple) else interval

    @property
    def end_date(self) -> Optional[dt]:
        """End Date."""
        if self.datetime is None:
            return self.datetime
        interval = str_to_interval(self.datetime)
        return interval[1] if isinstance(interval, tuple) else None
```

Also, is `None` correct for end date, instead of just the single datetime? E.g. this would break with how things are currently written, but seems like a reasonable thing to do?

```python
query = ItemCollectionUri(datetime="2025-01-14T16:00:00Z")
if query.start_date >= item.datetime and query.end_date <= item.datetime:
   ...
```

stac_fastapi/types/stac_fastapi/types/search.py
```diff
@@ -170,16 +152,47 @@ class BaseSearchGetRequest(APIRequest):
...
+        return interval if isinstance(interval, dt) else interval[0]
+
+    @property
+    def end_date(self) -> Optional[dt]:
+        """End Date."""
+        if self.datetime is None:
+            return self.datetime
+        interval = str_to_interval(self.datetime)
+        return interval[1] if isinstance(interval, tuple) else None
+
```
gadomski, line 195: It's a shame this is duplicated from `ItemCollectionUri` ... though b/c **stac-fastapi** is spaghetti-enough, it's probably not worth adding a mixin, so it's fine.

vincentsarago: yeah I tried to create a `mixin` class but didn't properly worked 

Let me retry

stac_fastapi/api/stac_fastapi/api/models.py
vincentsarago: good catch, will add some tests 🙏

vincentsarago: > Also, is None correct for end date, instead of just the single datetime?

maybe we could add another property `is_interval()`

vincentsarago: FYI: we don't use start/end dates in stac-fastapi-pgstac, I've added those to match with https://github.com/stac-utils/stac-pydantic/blob/main/stac_pydantic/api/search.py#L51-L58 which will also return `None` for a single datetime 

maybe we don't really need this anyway

gadomski: Yeah, I think an `.contains_datetime` method that handles the interval and non-interval cases would be safter than exposing derived `start_datetime` and `end_datetime`.

vincentsarago: what should `contains_datetime` return ?

gadomski: `bool | None` I think?

```python
def contains_datetime(self, datetime: dt) -> bool | None:
    if self.datetime is None:
        return None
    if self.datetime_is_interval():
        start, end = self.split_datetime()  # psuedocode
        return datetime >= start and datetime <= end
    else:
        return self.datetime == datetime
```

As I write this, just wanted to check we're handling no-timezone-datetimes correctly too (i.e. forcing them to UTC)?

gadomski (approved):
KISS 👍🏼

stac_fastapi/types/stac_fastapi/types/search.py
```diff
@@ -170,9 +201,7 @@ class BaseSearchGetRequest(APIRequest):
             },
         ),
     ] = attr.ib(default=None)
-    datetime: Optional[DateTimeType] = attr.ib(
-        default=None, converter=_datetime_converter
-    )
+    datetime: DateTimeQueryType = attr.ib(default=None, validator=_validate_datetime)
```
vincentsarago, line 204: Technically we didn't had to add `datetime` attribute here, but we do it to keep the `order` of the attributes the same as before (mostly for openapi docs 🤷‍♂️)

Issue #754, opened 2024-09-24
rename `filter` to `filter_body` for internals 


vincentsarago:
https://github.com/stac-utils/stac-fastapi/blob/753341f18d6c3d0454efef52d0ff841cfa3215f7/stac_fastapi/extensions/stac_fastapi/extensions/core/filter/request.py#L19-L29

In ☝️ we use `filter` which internally can conflict with the python `filter` method 

```python
@attr.s
class FilterExtensionGetRequest(APIRequest):
    """Filter extension GET request model."""

    filter_body: Annotated[
        Optional[str],
        Query(
            alias="filter",
            description="""A CQL filter expression for filtering items.\n
Supports `CQL-JSON` as defined in https://portal.ogc.org/files/96288\n
Remember to URL encode the CQL-JSON if using GET""",
            json_schema_extra={
                "example": "id='LC08_L1TP_060247_20180905_20180912_01_T1_L1TP' AND collection='landsat8_l1tp'",  # noqa: E501
            },
        ),
    ] = attr.ib(default=None)
    ...


class FilterExtensionPostRequest(BaseModel):
    """Filter extension POST request model."""

    filter_body: Optional[Dict[str, Any]] = Field(
        default=None,
        alias="filter",
        description="A CQL filter expression for filtering items.",
        json_schema_extra={
            "example": {
                "op": "and",
                "args": [
                    {
                        "op": "=",
                        "args": [
                            {"property": "id"},
                            "LC08_L1TP_060247_20180905_20180912_01_T1_L1TP",
                        ],
                    },
                    {"op": "=", "args": [{"property": "collection"}, "landsat8_l1tp"]},
                ],
            },
        },
    )


```

PR #784, opened 2025-01-17
rename `filter` to `filter_expr` to avoid python method conflict

vincentsarago:
closes https://github.com/stac-utils/stac-fastapi/issues/754


stac_fastapi/extensions/tests/test_collection_search.py
```diff
@@ -185,7 +185,7 @@ def test_collection_search_extension_models():
         assert "datetime" in response_dict
         assert "limit" in response_dict
         assert "q" in response_dict
-        assert "filter" in response_dict
+        assert "filter_expr" in response_dict
```
gadomski, line 188: Should we really be returning `filter_expr` from the `collections` endpoint? I would think we'd want to use the alias here.

gadomski (approved):

PR #785, opened 2025-01-17
Release/v3.1.0

vincentsarago:


gadomski (approved):

stac_fastapi/api/setup.py
```diff
@@ -7,7 +7,7 @@
 
 install_requires = [
     "brotli_asgi",
-    "stac-fastapi.types~=3.0",
+    "stac-fastapi.types~=3.1",
```
vincentsarago, line 10: @gadomski I'd love to have your objective opinion on this

Because 3.1 introduces  breaking changes (filter -> filter_expr and datetime type) should we:

- make a 4.0 instead
- set `api` and `extensions` dependencies to `>=3.1,<4.0` 
- set `api` and `extensions` dependencies to `==3.1.*` 
- 
ref: https://github.com/stac-utils/stac-fastapi/issues/678

gadomski: 🤔 tricky one. This repo is like halfway between a library and an application (in my head) which makes versioning weird. I'm a bit more inclined to just go for a 4.0 even though there isn't too much in it ... versions are cheap and at least we'll surprise other folks _less_ when their code breaks?

vincentsarago: well, looking at https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/blob/main/CHANGELOG.md makes me also think we should do 4.0 🙏

PR #798, opened 2025-02-06
remove child links (collections) in landing page response

### LINKED ISSUE #797: [Proposal] Remove `Child` links (collections) in Landing page
By default we add all the `collections` links to the landing page, but when the catalog grows this might become useless or really slow (we have catalogs with > 100 collections). 

https://github.com/stac-utils/stac-fastapi/blob/c71a355c46e53af40cd9ebe23bc151fa9faf8776/stac_fastapi/types/stac_fastapi/types/core.py#L418-L429

Note: When using pgstac with the `collection-search` extension enabled, we now return a default of 10 collections (default limit in pgstac) so the landing page of stac-fastapi-pgstac will only return the first 10 collections links 

@m-mohr not sure if stac-browser is using the landing page child links or directly the `/collections` link

vincentsarago:
closes #797

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/core.py
+++ b/stac_fastapi/types/stac_fastapi/types/core.py
@@ -415,19 +415,6 @@ class BaseCoreClient(LandingPageMixin, abc.ABC):
                 ]
             )
 
-        # Add Collections links
-        collections = self.all_collections(request=kwargs["request"])
-
-        for collection in collections["collections"]:
-            landing_page["links"].append(
-                {
-                    "rel": Relations.child.value,
-                    "type": MimeTypes.json.value,
-                    "title": collection.get("title") or collection.get("id"),
-                    "href": urljoin(base_url, f"collections/{collection['id']}"),
-                }
-            )
-
         # Add OpenAPI URL
         landing_page["links"].append(
             {
@@ -637,19 +624,6 @@ class AsyncBaseCoreClient(LandingPageMixin, abc.ABC):
                         "href": urljoin(base_url, "aggregations"),
                     },
                 ]
-            )
-
-        # Add Collections links
-        collections = await self.all_collections(request=kwargs["request"])
-
-        for collection in collections["collections"]:
-            landing_page["links"].append(
-                {
-                    "rel": Relations.child.value,
-                    "type": MimeTypes.json.value,
-                    "title": collection.get("title") or collection.get("id"),
-                    "href": urljoin(base_url, f"collections/{collection['id']}"),
-                }
             )
 
         # Add OpenAPI URL
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+
+### Changed
+
+- remove `child` links (`collections`) in landing page response
+
 ## [5.0.2] - 2025-01-30
 
 ### Fixed


stac_fastapi/types/stac_fastapi/types/core.py
```diff
@@ -415,19 +415,6 @@ def landing_page(self, **kwargs) -> stac.LandingPage:
...
-        for collection in collections["collections"]:
-            landing_page["links"].append(
-                {
-                    "rel": Relations.child.value,
-                    "type": MimeTypes.json.value,
-                    "title": collection.get("title") or collection.get("id"),
-                    "href": urljoin(base_url, f"collections/{collection['id']}"),
-                }
-            )
-
```
gadomski, line 430: How much extra complexity does it add to make this configurable? Smaller catalogs might want to keep these links — I know that I like seeing them when I hit an unfamiliar API, since the `/collections` endpoint can be kinda huge.

vincentsarago: what we could do is to not show the collections if the `collection-search` extension is enables 🤷‍♂️

vincentsarago: I should also point out that removing the `all_collections` call makes it easier to customize the all_collections method https://github.com/developmentseed/eoapi-devseed/pull/17/files#diff-035276bc260c2e34aa0886659327fe1950d424a2342a58879aa180f4e62e4d77R210-R223

gadomski: As a user, those two things aren't obviously related — that feels like a **pgstac** implementation detail leaking into this general API framework library?

gadomski: > with the assumption that smaller catalog do not need collection-search 😅

That's a fair point. I think as long as it was documented it'd make sense.

m-mohr: The childs added to the landing page is kinda inconsistent anyway, or does the implementation also add all the item links to the collections, too? As the latter is not done (for good reasons), the former should also not be done by default. It's a somewhat weird mix of API and static for an unknown usecase ;-)

vincentsarago: >  that feels like a pgstac implementation detail leaking into this general API framework library?

I agree but at the same time if your catalog is big (whatever your backend) the landing page will have way too much link. 

I feel will be really easy to customize your client to add back the links if you want them

```python
import attr
from fastapi import Request
from stac_fastapi.pgstac.core import CoreCrudClient
from stac_fastapi.types.stac import LandingPage

@attr.s
class CustomClient(CoreCrudClient):

    async def landing_page(
        self,
        request: Request,
        **kwargs,
    ) -> LandingPage:
        landing_page = super().landing_page(request, **kwargs)

        collections = self.all_collections(request=kwargs["request"])

        for collection in collections["collections"]:
            landing_page["links"].append(
                {
                    "rel": Relations.child.value,
                    "type": MimeTypes.json.value,
                    "title": collection.get("title") or collection.get("id"),
                    "href": urljoin(base_url, f"collections/{collection['id']}"),
                }
            )

        return landing_page
```

While if we keep it like it is right now, to `remove` the call to the `all_collections()` method, implementer need to re-write the whole method 🤷

vincentsarago: so I'm not sure what we should do ... 

IMO 
- make it configurable will add complexity to stac-fastapi
- removing it enable customization of the `all_collections` method 
- keep it as it is now can be annoying for large catalog

gadomski: > removing it enable customization of the all_collections method

👍🏼

PR #817, opened 2025-04-11
add enable_direct_response option

vincentsarago:
ref https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch/issues/347

cc @Zaczero @jonhealy1

### DIFF
--- a/stac_fastapi/types/stac_fastapi/types/config.py
+++ b/stac_fastapi/types/stac_fastapi/types/config.py
@@ -2,7 +2,9 @@
 
 from typing import Optional
 
+from pydantic import model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
+from typing_extensions import Self
 
 
 class ApiSettings(BaseSettings):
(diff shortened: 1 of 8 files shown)


stac_fastapi/types/stac_fastapi/types/config.py
```diff
@@ -27,14 +29,29 @@ class ApiSettings(BaseSettings):
     app_host: str = "0.0.0.0"
     app_port: int = 8000
     reload: bool = True
+
+    # Enable Pydantic validation for output Response
     enable_response_models: bool = False
 
+    # Enable direct `Response` from endpoint, skipping validation and serialization
+    enable_direct_response: bool = False
```
vincentsarago, line 37: any opinions on the option's name?

jonhealy1: LGTM

jonhealy1: You don't want to enable direct response by default? It's a new change, better to have it false.

jonhealy1 (approved):
Looks great. Thanks for this @vincentsarago

stac_fastapi/types/stac_fastapi/types/config.py
jonhealy1: True makes sense

PR #825, opened 2025-04-17
add health endpoint

### LINKED ISSUE #824: Add `/_mgmt/health` endpoint
Right now when we create an application we register a `/_mgmt/ping` endpoint to enable check for `liveliness`. This if fine but usually system (e.g Kubernetes) would want to check for `readiness` (a.k.a make sure database connection works, ...) 

https://github.com/stac-utils/stac-fastapi/blob/a402be81b0480ef500db598703c89b10417e57e8/stac_fastapi/api/stac_fastapi/api/app.py#L364-L373

I propose we register another `health` endpoint `/_mgmt/health` which can return more information about the application `readiness`.

ref: https://github.com/stac-utils/stac-fastapi-pgstac/issues/229

cc @emmanuelmathot

vincentsarago:
closes #824

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -132,7 +132,7 @@ class StacApi:
         )
     )
     route_dependencies: List[Tuple[List[Scope], List[Depends]]] = attr.ib(default=[])
-    health_check: Union[Callable[[], [Dict]], Callable[[], Awaitable[Dict]]] = attr.ib(
+    health_check: Union[Callable[[], Dict], Callable[[], Awaitable[Dict]]] = attr.ib(
         default=lambda: {"status": "UP"}
     )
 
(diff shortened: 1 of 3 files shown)


stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -363,14 +370,44 @@ def register_core(self) -> None:
...
+                    "content": {
+                        MimeTypes.json.value: {},
+                    },
+                },
+            },
+            response_class=self.response_class,
+            methods=["GET"],
+            endpoint=ping,
+            tags=["Liveliness/Readiness"],
+        )
```
vincentsarago, line 393: I don't think we needed to create a `mgmt_router`. I changed the logic to match what we do for the other endpoints (using `router.add_api_route`)

gadomski (approved):

PR #827, opened 2025-04-17
revert to custom router for mgmt routes

vincentsarago:
this https://github.com/stac-utils/stac-fastapi/pull/825/files#r2048617917 was wrong!


gadomski (approved):
👍🏼 apologies for not catching the first time 🙇🏼

PR #842, opened 2025-05-19
add type checking

### LINKED ISSUE #828: Improve Type Hints
ref #826 

in https://github.com/stac-utils/stac-fastapi/blob/main/CONTRIBUTING.md we says we run mypy during pre-commit but it's not true. The lib is not type checked 😓

We also don't provided `typed.py` in the sub-module.

vincentsarago:
same as https://github.com/stac-utils/stac-fastapi/pull/841 but without breaking changes 😅 

closes #828

### DIFF
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -110,7 +110,7 @@ class StacApi:
             ),
             takes_self=True,
         ),
-        converter=update_openapi,
+        converter=update_openapi,  # type: ignore
     )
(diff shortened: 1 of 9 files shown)


stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -110,7 +110,7 @@ class StacApi:
             ),
             takes_self=True,
         ),
-        converter=update_openapi,
+        converter=update_openapi,  # type: ignore
```
vincentsarago, line 113: for some reason `attr + Mypy` doesn't like this (if we ever move to `attrs.field` it works fine

gadomski (approved):

PR #655, opened 2024-04-10
Adding free-text extension.

rhysrevans3:
**Related Issue(s):**

- #

**Description:**

Adding the [free-text extension](https://github.com/stac-api-extensions/freetext-search)

**PR Checklist:**

- [x] `pre-commit` hooks pass locally
- [x] Tests pass (run `make test`)
- [ ] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [ ] Changes are added to the [CHANGELOG](https://github.com/stac-utils/stac-fastapi/blob/main/CHANGES.md).

PR #849, opened 2025-07-07
fix extension free-text advanced GET-q query not split by comma

fmigneault:
**Related Issue(s):**

- n/a

**Description:**

Unlike its *basic* counterpart, the `q` parameter of *advanced* free-text extension was passed as is (plain string) to the downstream client. This lead to invalid searches.

image

This PR fixes this by applying the same transform, which leads to the expected behavior.

image

>[!NOTE]
> Both "test_search_free_text_advanced" tests fail currently, but it doesn't seem to make much sense that the literal inputs are returned as is in `response.json() == "+ocean,-coast"`, given that the client expects `q: List[str]` (https://github.com/stac-utils/stac-fastapi-pgstac/blob/d33b0f3d1d509e16fe0dfc765dc47aea6613aac8/stac_fastapi/pgstac/core.py#L57). Should those tests be adjusted?

```python
stac_fastapi/extensions/tests/test_free_text.py:302: AssertionError
======================================================================================================================================= short test summary info =======================================================================================================================================
FAILED stac_fastapi/extensions/tests/test_free_text.py::test_search_free_text_search_advanced - AssertionError: assert ['+ocean', '-coast'] == '+ocean,-coast'
FAILED stac_fastapi/extensions/tests/test_free_text.py::test_search_free_text_advanced_complete - AssertionError: assert ['ocean', 'coast'] == 'ocean,coast'
```

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- fix `q` query parameter parsing of advanced free-text search on GET request
+
 ## [6.0.0] - 2025-06-19
 
 ### Changed
--- a/stac_fastapi/extensions/stac_fastapi/extensions/core/free_text/request.py
+++ b/stac_fastapi/extensions/stac_fastapi/extensions/core/free_text/request.py
@@ -48,7 +48,7 @@ class FreeTextAdvancedExtensionGetRequest(APIRequest):
     """Free-text Extension GET request model."""
 
     q: Annotated[
-        Optional[str],
+        Optional[List[str]],
         Query(
             description="Parameter to perform free-text queries against STAC metadata",
             openapi_examples={
@@ -56,7 +56,7 @@ class FreeTextAdvancedExtensionGetRequest(APIRequest):
                 "Coastal": {"value": "ocean,coast"},
             },
         ),
-    ] = attr.ib(default=None)
+    ] = attr.ib(default=None, converter=_ft_converter)
 
 
 class FreeTextAdvancedExtensionPostRequest(BaseModel):


vincentsarago:
This was made on purpose see https://github.com/stac-utils/stac-fastapi/pull/655#issuecomment-2236300851

advanced might get removed anyway https://github.com/stac-api-extensions/freetext-search/pull/13

fmigneault:
@vincentsarago
Thanks for the heads up! 
Is this removal sure to make it in? I will refer to basic free-text in our implementation if so.

What would be the recommended approach to do the `+/-` of [Advanced FreeText-Search](https://github.com/stac-api-extensions/freetext-search?tab=readme-ov-file#advanced) instead? For users, this seemed like a convenient way to filter-out items without involving "more complex" CQL2, which is why we were considering enabling it.

vincentsarago:
> What would be the recommended approach to do the +/- of [Advanced FreeText-Search](https://github.com/stac-api-extensions/freetext-search?tab=readme-ov-file#advanced) instead? For users, this seemed like a convenient way to filter-out items without involving "more complex" CQL2, which is why we were considering enabling it.

I can see the appeal for `advanced` free-text but this is lacking an official (and complete) specification (on OGC or STAC side)

going back to the original issue, in `stac-fastapi-pgstac` we use string anyway https://github.com/stac-utils/stac-fastapi-pgstac/blob/main/stac_fastapi/pgstac/core.py#L598-L599 👀 


FYI: I also think that free-text is not recommended for items search in pgstac

fmigneault:
@rhysrevans3 
Since there is a `q` parameter conflict between the basic/advanced, I think it is actually clearer if both are together as currently defined. This makes it clear that implementer have to pick one or the other, and users can better understand their differences. I think that moving it will increase the chances of user being unaware of the advanced variant if they land only on the basic description, or they might get lost amongst all other STAC extensions and conformance links.

@vincentsarago 
Yes, that works also. 
I think part of the issue is that it is not clear at which level the parsing should occur right now.
Currently, comma-word splitting happens while validating the parameter, but `OR` joining happens in `pgstac`. If the `str` is returned directly, does that mean the comma-word split would happen in `pgstac` as well for basic search, or that means `pgstac` will deal with `List[str] | str` directly for basic/advanced respectively?

vincentsarago:
FYI: https://github.com/stac-utils/stac-fastapi-pgstac/pull/263

if/when merged, I'll back port the change and release a `5.0.3` with this change

fmigneault:
@vincentsarago 
Seems to work correctly with https://github.com/stac-utils/stac-fastapi/releases/tag/6.0.0
Thanks

PR #879, opened 2026-02-07
fix broken starlette cors middleware signature 

fmigneault:
**Related Issue(s):**

- https://github.com/Kludex/starlette/pull/3065

**Description:**

In `starlette==0.51.0`, the `CORSMiddleware` adds a new parameter, which breaks the current `stac-fastapi` implementation that does not pin higher version, due to unpinned `fastapi` also no limiting the upper version.

The problem can be observed when using the API with mismatching `starlette` versions, where the `600` default of `max_age` parameter ends up making its way into `expose_headers` because of the parameter offset, leading to an error with `", ".join(expose_headers)`.

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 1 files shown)


fmigneault:
@vincentsarago
Let me know if this fix is sufficient or another alternative should be taken.

PR #880, opened 2026-02-07
Multi-tenant /Catalogs Extension

jonhealy1:
**Related Issue(s):**

- #886 

### Description: Multi-Tenant Catalogs Extension

This extension introduces a recursive `/catalogs` endpoint to the STAC API, enabling complex, nested hierarchies beyond the standard flat `Root -> Collections` structure. It transforms a STAC API into a Multi-Tenant system capable of serving distinct catalog trees (e.g., `Provider -> Theme -> Project`) within a single instance.

**Key Architectural Concepts:**

* **Recursive Hierarchy:** Unlike standard STAC, which flattens data into a single list of Collections, this extension allows Catalogs to contain other Sub-Catalogs to unlimited depth.
* **Poly-Hierarchy (Virtual Organization):** Collections are not physically moved but logically linked. A single Collection can belong to multiple parent Catalogs simultaneously (e.g., a "Sentinel-2" collection can exist under both a "USGS" catalog and an "Optical Data" catalog).
* **Safety-First Management:** The architecture strictly separates "Organization" from "Data."
    * **Unlinking vs. Deleting:** Deleting a Catalog via this extension is non-destructive to the actual data. It "unlinks" the child Collections.
    * **Orphan Adoption:** If a Collection is unlinked from its last parent, it is automatically adopted by the Root Catalog to ensure no data is ever lost or becomes undiscoverable.
* **Unified Discovery:** Integrates with the **Children Extension** to provide a single view (`/children`) that lists both Sub-Catalogs and Collections, supporting optional type filtering.
* **Configurable Transactions.** The extension supports an enable_transactions flag (default: False).

**Full Endpoint List:**

* **Registry & Root:**
    * `GET /catalogs` - List all root catalogs.
    * `POST /catalogs` - Register a new root catalog.
* **Catalog Management:**
    * `GET /catalogs/{catalog_id}` - Get catalog metadata.
    * `PUT /catalogs/{catalog_id}` - Update catalog metadata.
    * `DELETE /catalogs/{catalog_id}` - Disband a catalog (does not delete data).
* **Sub-Catalogs (Recursive):**
    * `GET /catalogs/{catalog_id}/catalogs` - List sub-catalogs.
    * `POST /catalogs/{catalog_id}/catalogs` - Create or link a sub-catalog.
    * `DELETE /catalogs/{catalog_id}/catalogs/{sub_catalog_id}` - Unlink a sub-catalog.
* **Collections (Poly-hierarchy):**
    * `GET /catalogs/{catalog_id}/collections` - List linked collections.
    * `POST /catalogs/{catalog_id}/collections` - Create or link a collection.
    * `GET /catalogs/{catalog_id}/collections/{collection_id}` - Get collection details.
    * `DELETE /catalogs/{catalog_id}/collections/{collection_id}` - Unlink a collection.
* **Items (Scoped Access):**
    * `GET /catalogs/{catalog_id}/collections/{collection_id}/items` - Search items within catalog context.
    * `GET /catalogs/{catalog_id}/collections/{collection_id}/items/{item_id}` - Get single item.
* **Discovery & Capabilities:**
    * `GET /catalogs/{catalog_id}/children` - Unified list of child Catalogs and Collections.
    * `GET /catalogs/{catalog_id}/conformance` - Conformance classes for this catalog tree.
    * `GET /catalogs/{catalog_id}/queryables` - Queryable fields for this catalog tree.

**Specification Reference:**
[Healy-Hyperspatial/multi-tenant-catalogs](https://github.com/Healy-Hyperspatial/multi-tenant-catalogs)

**PR Checklist:**


jonhealy1:
@vincentsarago I agree. When and if this extension gains more traction, I would definitely be open to moving it out of our repository.

hrodmn (commented):
Thanks for all of the work on this @jonhealy1!

I have a few ideas on some structural choices that might make this extension work better for more users and make better use of existing extensions. Mainly I think these ideas boil down to adding some logic to a class definition method like we have for the `CollectionSearch` extension that defines request models based on the other extensions that are enabled alongside it.

https://github.com/stac-utils/stac-fastapi/blob/600b653c4f94357b62a3f526fbdab3e73d9d96b8/stac_fastapi/extensions/stac_fastapi/extensions/core/collection_search/collection_search.py#L74-L104

The advantage to doing this is that the API documentation will reflect subtle changes to the available parameters for each endpoint, e.g. if `CatalogsExtension` and `FilterExtension` are enabled, the `filter` parameter would be added to the `/catalogs` endpoint.

Most importantly, I would want some way to opt in/out of the transactions aspect of this extension, and I would want the default behavior for catalog transactions to be OFF. For example, there may be users who want to enable GET requests but not POST, PUT or DELETE because they have some other system for managing i/o on their STAC and they do not want to manage auth on the public API. If the `CatalogsExtension` were aware of the `TransactionExtension` it could only switch on the CRUD endpoints if the `TransactionExtension` is enabled.

stac_fastapi/extensions/stac_fastapi/extensions/third_party/multi_tenant_catalogs/catalogs.py
```diff
@@ -0,0 +1,447 @@
...
+        )
+
+    async def _unlink_sub_catalog_wrapper(
+        self, catalog_id: str, sub_catalog_id: str, request: Request
+    ) -> None:
+        return await self.client.unlink_sub_catalog(
+            catalog_id=catalog_id, sub_catalog_id=sub_catalog_id, request=request
+        )
+
+    def register(self, app: FastAPI, settings: Optional[Dict[str, Any]] = None) -> None:
```
gadomski: Superclass doesn't have `settings` in the signature.

```suggestion
    def register(self, app: FastAPI) -> None:
```

gadomski (dismissed):
I did more of a nitpick code review rather than an overall architecture review. High-level ask, I know @vincentsarago asked for the move, but any way we could do the `bulk_transaction` move in its own PR?

jonhealy1:
Hi @bkanuka, thanks for checking this out and for the great questions!

To give you a little behind-the-scenes context: we actually originally pitched this extension as the **"Virtual Catalogs Endpoint"**. We really see the number one innovation here as introducing the `/catalogs` endpoint natively into the dynamic API layer - transforming the STAC API from a flat list of `/collections` into a navigable, hierarchical system. The goal was to support flexible "playlists" or semantic themes (like a SKOS registry) where a single Collection could logically exist in multiple different folders simultaneously. However, the STAC Project Steering Committee (PSC) felt "Multi-Tenant" was a more appropriate term, so we adopted their recommendation. It definitely might create some confusion in the future.

Here is the easiest way to think about the difference between the VEDA PR and this extension:

**1. VEDA’s Approach: Access-Level Isolation**
The VEDA PR solves multi-tenancy at the **routing and filtering** level. 
* It takes a standard STAC API (which relies on a single, flat list of `/collections`) and uses API middleware to intercept a `{tenant}` prefix in the URL.
* It then silently injects filters into the database queries so that "Tenant A" only sees their specific slice of that flat Collections list. 
* *Analogy:* It's like having one big filing cabinet, but giving different users magic glasses so they only see their own files.

**2. This Extension's Approach: Structural Namespaces**
This extension solves multi-tenancy at the **database and structural** level.
* **Crucially, this can allow you to natively store and serve multiple distinct Catalogs on the exact same database - a capability that is new to STAC APIs.**
* Instead of keeping a flat list of collections, it introduces recursive STAC Catalogs (essentially, folders) directly into the API and database.
* You can build a deeply nested, navigable tree (e.g., `/catalogs/usgs/catalogs/landsat/collections/...`). 
* Because it's a structural **poly-hierarchy**, a single Collection can logically exist inside multiple different Catalogs simultaneously (e.g., in a "USGS" catalog and a "Forestry" catalog) without duplicating the underlying data.
* **Built-in Data Safety:** Organizing your STAC API with this extension is literally like creating a music playlist. The `/catalogs` transaction endpoints strictly manage *links*, not data. You can safely build, rearrange, or delete entire sub-catalogs, and it will only "unlink" the Collections. Actual data destruction is intentionally restricted to the normal STAC API transaction routes (e.g., `DELETE /collections/{id}`).
* *Analogy:* It's like building actual, labeled drawers and folders inside the filing cabinet, allowing users to physically browse the organizational tree without ever risking the files themselves.

**The Goal & `pgstac`**
The goal isn't to force a single architectural choice. In fact, these two approaches aren't mutually exclusive- you could absolutely have structural catalog folders *and* strict access-control filters! But our primary goal here was to formalize a standard, spec-compliant way to navigate massive STAC APIs that have outgrown the flat `/collections` list.

We currently have a working implementation operating as an optional feature in our [stac-fastapi-elasticsearch-opensearch](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch) backend. 

Given how powerful `pgstac`'s relational model is, I think it is the absolute perfect candidate for this. Adding a `catalogs` table and a many-to-many relationship for collections would be incredibly powerful for the community.

jonhealy1:
> Thanks for the thorough reply. It clarified a few things. I like the flexibility of this many-to-many approach and hopefully it gets wider adoption. (Although the route-to-filter approach does have the benefit of being a bit "thinner").
> 
> I'll also take a look at the electric/open search implementation.
> 
> Is there a technical reason this was first implemented in opensearch over pgstac (i.e. there's something that makes opensearch a better backend for this) or was it a client-driven decision?

@bkanuka You are totally right that the route-to-filter approach is "thinner" - it's a very elegant solution if the only goal is access control! It doesn't accomplish what we set out to achieve here, namely structural, nested Catalogs that allow for things like SKOS and improved discoverability.

To answer your question: there is no technical reason it was implemented in OpenSearch first, and OS isn't necessarily a "better" backend for this. It was simply because I am the lead maintainer of the `stac-fastapi-elasticsearch-opensearch` project. We have had people asking for a `/catalogs` endpoint for years, and we are currently talking to an outside organization who is very interested in this approach for SKOS.

Other organizations run their own custom versions of this already, but it's currently a "wild west" scenario because a `/catalogs` route has never been officially spec'ed out and documented.

In our OpenSearch implementation, we define Catalogs as a specific type of Collection. This meant we didn't even have to add new indexes (tables) or introduce breaking changes to support the functionality.

hrodmn (commented):
I realize that it is inherently backend-specific, but it would be good to note somewhere that, since links tend to be dynamically generated by stac-fastapi applications, the relationships between catalogs and collections (and catalogs and catalogs) needs to be tracked somehow.

I am also wondering if the `children` extension gets activated does that mean the API root will potentially have a `rel=child` link for all collections (not just sub-catalogs)?

bitner:
so the thought i just had. could this be implemented backend agnostic by just acting as a proxy to rewrite queries inro collection search requests similar to how stac-auth-proxy works rather than needing to implement in stac-fastapi-pgstac at all??

jonhealy1:
> so the thought i just had. could this be implemented backend agnostic by just acting as a proxy to rewrite queries inro collection search requests similar to how stac-auth-proxy works rather than needing to implement in stac-fastapi-pgstac at all??

@bitner It's an interesting idea. Some quick thoughts:

1. You would need to store the DAG structure somewhere. This extension allows complex catalog and sub-catalog nesting. The initial motivation for this was to help provide SKOS thematic organization for large providers like ESA and NASA. There may be performance issues with searching a separately stored data structure - "give me every id in this sub-tree" - and then sending large api requests to the backend.

2. This extension provides a virtual, catalog DAG management system via catalog transactions that would be difficult to emulate via a proxy. It could be done via the externally stored DAG, but one issue would be state synchronization: let's say you delete a collection via the normal STAC API routes - how would your proxy's DAG know it was deleted?

3. We are looking at adding sub-catalog search soon and while this could be done via proxy- find the collections and catalogs under a sub-tree and do collections-search - similarly to point 1, proxy API requests would get very large and could hit URL length or payload limits.

4. Another feature we would like to add is the ability to create virtual collections of items based on cql2 filtering. Let's say you want a virtual collection based on geography. A proxy would have to parse and inject complex CQL2 on the fly.

5. Dynamic Link Rewriting: STAC relies heavily on relational links (`self`, `parent`, `root`, `child`). If a proxy is creating a virtual structure, it must intercept every JSON response from the backend, parse it, rewrite all the `href` URLs to match the virtual proxy paths, and repackage it. This adds significant latency and computational overhead to every single request.

6. Pagination Complexity: If a proxy is faking a sub-catalog search by querying multiple underlying collections, managing pagination becomes a nightmare. The proxy would have to track and merge multiple backend continuation tokens (cursors/offsets) to generate a single valid `next` link for the client.

7. Double API Latency & Payload Size: By introducing a proxy, we are forcing every request to travel through two complete HTTP layers.

I think the complexity of creating a proxy that added all of the features our extension provides - and more in the future -combined with the performance issues and JSON-rewriting overhead would make me think it's not the best solution.

jonhealy1:
Reopening. Thoughts:

- All of the other `stac-fastapi` extensions live here
- How will the community know about this extension if it's hosted somewhere else and we don't want to add a link to the extension in the `Readme`
- This extension, like all other extensions, are optional
- Similarly, the code from this extension does not touch any of the core api code - there is no danger to approving this pr
- This extension is being used in [SFEOS](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch), the second most popular stac-fastapi implementation and an important project to many organisations. 
- There is a draft [pr](https://github.com/stac-utils/stac-fastapi-pgstac/pull/366) open in stac-fastapi-pgstac to implement this extension there
- People have been asking for a /catalogs route for years
- Our extension is not closed source and is open to new ideas and input

@gadomski has said that there isn't a consensus towards eventually approving this pr or not. I would like to know what the objections are. Has any other extension ever been refused in this project? Some extensions hosted here don't even have tests https://github.com/stac-utils/stac-fastapi/issues/901

jonhealy1:
@hrodmn I have some time coming up to clean up the docstrings and divide this into two separate extensions. Thanks for all of the input

jonhealy1:
Slowly adding to stac-fastapi-pgstac: https://github.com/stac-utils/stac-fastapi-pgstac/pull/366

vincentsarago:
I've updated this branch but I think this now lives in https://github.com/StacLabs/stac-fastapi-catalogs-extension

jonhealy1:
> I've updated this branch but I think this now lives in https://github.com/StacLabs/stac-fastapi-catalogs-extension

This was open a for a while so I moved the extension there. Then there were opinions about where the extension should rightfully live. I'm good with either option.

jonhealy1:
Hosted here: https://github.com/StacLabs/stac-fastapi-catalogs-extension

PR #885, opened 2026-02-16
feat: add route_dependencies attribute to transaction extensions

### LINKED ISSUE #884: add optional router dependencies for TransactionExtension and BulkTransactionsExtension to facilitate adding authentication to those endpoints
It is currently possible to add dependencies to a `StacApi` with the `add_route_dependencies` method, as shown in this test:

https://github.com/stac-utils/stac-fastapi/blob/2e2b0f9e65b78a07f6f9f68349372daa334779ac/stac_fastapi/api/tests/test_api.py#L118-L145

This works but it is **tricky** because you need to enumerate all of the route paths/methods that you want to cover. Would it make sense to add an optional `route_dependencies` attribute to the `TransactionsExtension` and `BulkTransactionExtension` that would be included in all of the `router.add_api_route` calls?

If we did this, then a user could easily define some kind of auth function and apply it as a dependency to all of the routes that are managed by the transaction extensions.

It could look something like this:
```python
from fastapi import Depends
from fastapi.security import HTTPBasic, HTTPBasicCredentials

from stac_fastapi.extensions.core import TransactionExtension


def must_be_bob(
    credentials: HTTPBasicCredentials = Depends(HTTPBasic()),
):
    if credentials.username == "bob":
        return True

    raise HTTPException(
        status_code=status.HTTP_401_UNAUTHORIZED,
        detail="You're not Bob",
        headers={"WWW-Authenticate": "Basic"},
    )


api = StacApi(
    ...
    extensions=[
        TransactionExtension(..., route_dependencies=[Depends(must_be_bob)])
    ],
    ...
)

```

On the `TransactionExtension` we would just add the new `route_dependencies` attribute and provide it in all calls to `self.router.add_api_route()`.

cc @vincentsarago @gadomski

hrodmn:
resolves #884

**Related Issue(s):**

- #884 

**Description:**

This adds a new argument to the `TransactionExtension` and `BulkTransactionExtension` that makes it simple to add a dependency to all routes affected by the transactions extensions. It is possible that we would want to build off of the existing `add_route_dependencies` pathway instead of doing this within the extension, see #884 for some discussion around that point.

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
 - Add tests for the `sort` extension ([#906](https://github.com/stac-utils/stac-fastapi/pull/906))
 - Add tests for the `fields` extension ([#902](https://github.com/stac-utils/stac-fastapi/pull/902))
 - Simply add a list of present stac-fastapi extensions to the Readme ([#898](https://github.com/stac-utils/stac-fastapi/pull/898))
-- Add `route_dependencies` argument to `transactions` extension to make it easy to apply auth dependencies to all transactions routes ([#885](https://github.com/stac-utils/stac-fastapi/pull/885))
+- Add `route_dependencies` argument to `transactions` and `bulk_transactions` extensions to make it easy to apply auth dependencies to all transactions routes ([#885](https://github.com/stac-utils/stac-fastapi/pull/885))
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


hrodmn:
@jonhealy1 @vincentsarago this is ready for review. I think it will be an important feature for users to safely enable the transactions extension in downstream applications like stac-fastapi-pgstac/SFEOS. Users who want more fine-grained control could still use the per-route approach that existed before but now there is an easy button to turn on authentication for all transactions routes!

jonhealy1 (approved):
Nice work! This makes adding dependencies to the transaction routes a lot easier.

jonhealy1:
This same idea should be added to the catalogs transaction extension as well

vincentsarago (approved):

Issue #881, opened 2026-02-09
move all extensions to a unique submodules


vincentsarago:
Ref https://github.com/stac-utils/stac-fastapi/pull/880#issuecomment-3871972570

Right now extensions are split into 2 sub-modules: `core` and `third-party`. There is no clear definition of what and why some extensions are or aren't `core`.

I personally think that we should move everything to the root directory

PR #899, opened 2026-04-02
Move all extensions to root

jonhealy1:
**Related Issue(s):**

- #881 

**Description:**

- removed the third party and core extensions folder 
- move all extensions to root
- this is a **Breaking Change** as the extension import paths will now change

**PR Checklist:**


jonhealy1:
Maybe these should be moved to root instead of being in the core folder?

stac_fastapi/extensions/stac_fastapi/extensions/core/__init__.py
```diff
@@ -1,39 +0,0 @@
-"""stac_api.extensions.core module."""
```
vincentsarago, line 1: I wonder if we could keep this and import all the extensions to keep backward compatibility. 

This is going to be an important breaking change

jonhealy1: Keep all the extensions in a core folder or just try to route them via the __init__. I can try.

jonhealy1: We probably don't want to release a new major version because of this :)

vincentsarago: > try to route them via the init. I can try.

I can have a look if you want

jonhealy1: Have a look if you want no problem

jonhealy1: Haha sorry got bored and figured something out

hrodmn (approved):
I think this makes sense as a non-breaking, backwards-compatible change. @vincentsarago is there anything else you would want to see before we merge this?

vincentsarago (approved):

PR #938, opened 2026-06-25
fix: avoid urljoin issue

vincentsarago:
Because we always add `/` at the end of the base url in 

https://github.com/stac-utils/stac-fastapi/blob/8ca103aca17d32cc65f7387043c3d78dbe372a05/stac_fastapi/types/stac_fastapi/types/requests.py#L12

```python
# when base url doesn't end with a slash, the last part of the base url is replaced by the relative url
urljoin("http://stac.io/stac", "collection")
'http://stac.io/collection'

urljoin("http://stac.io/stac/", "collection")
'http://stac.io/stac/collection'
```

There is only one place where the issue could happen in https://github.com/stac-utils/stac-fastapi/blob/8ca103aca17d32cc65f7387043c3d78dbe372a05/stac_fastapi/types/stac_fastapi/types/links.py#L21-L26

but I don't see any use of this method.

Anyway, better be safe than sorry so I decided to apply the same `fix` everywhere.

### DIFF
--- a/stac_fastapi/api/tests/test_app_prefix.py
+++ b/stac_fastapi/api/tests/test_app_prefix.py
@@ -88,7 +88,7 @@ def test_api_prefix(TestCoreClient, prefix):
             assert resp.status_code == 200
 
 
-@pytest.mark.parametrize("prefix", ["", "/a_prefix"])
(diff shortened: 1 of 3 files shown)


jonhealy1 (approved):
LGTM!

PR #945, opened 2026-06-30
Add sortables/ update sort extension to v1.1.0

jonhealy1:
**Related Issue(s):**

- #859 

**Description:**

- Added sortables support to Sort extension for discovering available sort fields via dedicated endpoints (`/sortables`, `/collections/{id}/sortables`, `/collections-sortables`) 
- Updated Sort extension to v1.1.0 specification with granular conformance classes for item search, OGC API Features, and collection search endpoints. Add `SearchSortExtension`, `ItemCollectionSortExtension`, and `CollectionSearchSortExtension` classes for context-specific sorting capabilities.

**PR Checklist:**

- [x] `pre-commit` hooks pass locally
- [x] Tests pass (run `make test`)
- [x] Documentation has been updated to reflect changes, if applicable, and docs build successfully (run `make docs`)
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/stac-fastapi/blob/main/CHANGES.md).

PR #954, opened 2026-07-21
fix: add back SortConformanceClasses enum values

vincentsarago:
@jonhealy1 I didn't realize that https://github.com/stac-utils/stac-fastapi/pull/945 had breaking changes. For now this PR will address the changes and add deprecation notice 

we can remove this in 7.0

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


jonhealy1 (approved):
Looks really good.  I should have caught this.

PR #958, opened 2026-07-22
Added optional prometheus metrics endpoint.

pantierra:
**Related Issue(s):**

- https://github.com/stac-utils/stac-fastapi-pgstac/pull/398

**Description:**

- Optional Prometheus metrics via `stac-fastapi-api[metrics]`, exposed at `{router_prefix}/_mgmt/metrics` when the extra is installed (same opt-in pattern as [sfeos](https://github.com/stac-utils/stac-fastapi-elasticsearch-opensearch)).
- Uses low-cardinality STAC `operation` labels (`search`, `get_item`, …) instead of default path-based labels, per the approach discussed in https://github.com/developmentseed/eoAPI/issues/193#issuecomment-4353287562.
- Backends can map custom routes with `register_operations`.

Concretely, things to discuss if this is desired:

- Without `[metrics]`, app still starts and no metrics route is registered
- With `[metrics]`, `/_mgmt/metrics` returns series labeled by `operation`
- Item/collection IDs do not appear in label values
- With a router prefix, metrics are at `{prefix}/_mgmt/metrics`

**PR Checklist:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dev = [
     "pre-commit",
     "httpx2",
     "requests",
-    "prometheus-fastapi-instrumentator>=8,<9",
+    "prometheus-fastapi-instrumentator>=8",
 ]
 docs = [
     "black>=23.10.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dev = [
     "pre-commit",
     "httpx2",
     "requests",
+    "prometheus-fastapi-instrumentator>=8,<9",
 ]
 docs = [
     "black>=23.10.1",
--- a/stac_fastapi/api/pyproject.toml
+++ b/stac_fastapi/api/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
 
 [project.optional-dependencies]
 metrics = [
-    "prometheus-fastapi-instrumentator>=8,<9",
+    "prometheus-fastapi-instrumentator>=8",
 ]
 
 [dependency-groups]
--- a/stac_fastapi/api/pyproject.toml
+++ b/stac_fastapi/api/pyproject.toml
@@ -35,6 +35,11 @@ dependencies = [
     "stac-fastapi.types~=6.4",
 ]
 
+[project.optional-dependencies]
+metrics = [
+    "prometheus-fastapi-instrumentator>=8,<9",
+]
+
 [dependency-groups]
 dev = [
     "httpx",
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -41,16 +41,6 @@ from stac_fastapi.types.extension import ApiExtension
 from stac_fastapi.types.search import BaseSearchGetRequest, BaseSearchPostRequest
 
 
-def add_metrics(app: FastAPI) -> None:
-    """Expose Prometheus metrics when the optional metrics extra is installed."""
-    try:
-        from stac_fastapi.api.metrics import instrument_app
-    except ImportError:
-        return
-
-    instrument_app(app)
-
-
 @attr.s
 class StacApi:
     """StacApi factory.
@@ -84,6 +74,10 @@ class StacApi:
         health_check:
             A Callable which return application's `health` information.
             Defaults to `def health: return {"status": "UP"}`
+        add_metrics:
+            When ``True``, instrument the app and expose Prometheus metrics at
+            ``{router_prefix}/_mgmt/metrics``. Requires
+            ``stac-fastapi-api[metrics]`` to be installed.
 
     """
 
@@ -152,6 +146,7 @@ class StacApi:
     health_check: Callable[[], dict] | Callable[[], Awaitable[dict]] = attr.ib(
         default=lambda: {"status": "UP"}
     )
+    add_metrics: bool = attr.ib(default=False)
 
     def get_extension(self, extension: type[ApiExtension]) -> ApiExtension | None:
         """Get an extension.
@@ -487,7 +482,10 @@ class StacApi:
         for middleware in self.middlewares:
             self.app.user_middleware.insert(0, middleware)
 
-        add_metrics(self.app)
+        if self.add_metrics:
+            from stac_fastapi.api.metrics import instrument_app
+
+            instrument_app(self.app)
 
         # customize route dependencies
         for scopes, dependencies in self.route_dependencies:
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -485,7 +485,10 @@ class StacApi:
         if self.add_metrics:
             from stac_fastapi.api.metrics import instrument_app
 
-            instrument_app(self.app)
+            instrument_app(
+                self.app,
+                endpoint=f"{self.app.state.router_prefix}/_mgmt/metrics",
+            )
 
         # customize route dependencies
         for scopes, dependencies in self.route_dependencies:
--- a/stac_fastapi/api/stac_fastapi/api/app.py
+++ b/stac_fastapi/api/stac_fastapi/api/app.py
@@ -41,6 +41,16 @@ from stac_fastapi.types.extension import ApiExtension
 from stac_fastapi.types.search import BaseSearchGetRequest, BaseSearchPostRequest
 
 
+def add_metrics(app: FastAPI) -> None:
+    """Expose Prometheus metrics when the optional metrics extra is installed."""
+    try:
+        from stac_fastapi.api.metrics import instrument_app
+    except ImportError:
+        return
+
+    instrument_app(app)
+
+
 @attr.s
 class StacApi:
     """StacApi factory.
@@ -476,6 +486,8 @@ class StacApi:
 
         for middleware in self.middlewares:
             self.app.user_middleware.insert(0, middleware)
+
+        add_metrics(self.app)
 
         # customize route dependencies
         for scopes, dependencies in self.route_dependencies:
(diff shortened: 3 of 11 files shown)


stac_fastapi/api/pyproject.toml
```diff
@@ -35,6 +35,11 @@ dependencies = [
     "stac-fastapi.types~=6.4",
 ]
 
+[project.optional-dependencies]
+metrics = [
+    "prometheus-fastapi-instrumentator>=8,<9",
```
gadomski: In libraries, I generally shy away from upper bounds on dependencies unless there's a good reason.

```suggestion
    "prometheus-fastapi-instrumentator>=8",
```

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dev = [
     "pre-commit",
     "httpx2",
     "requests",
+    "prometheus-fastapi-instrumentator>=8,<9",
```
gadomski:
```suggestion
    "prometheus-fastapi-instrumentator>=8",
```

stac_fastapi/api/stac_fastapi/api/app.py
```diff
@@ -477,6 +487,8 @@ def __attrs_post_init__(self) -> None:
         for middleware in self.middlewares:
             self.app.user_middleware.insert(0, middleware)
 
+        add_metrics(self.app)
+
```
gadomski: Rather than implicitly adding if the import succeeds, I _think_ it'd be better to make this opt-in (e.g. with an "add_metrics" argument) and then blow up if the prometheus import fails?

To put it another way, I think it'd be nicer to opt-in via code (`StacApi(add_metrics=True, ...)`) rather than via environment.

stac_fastapi/api/stac_fastapi/api/metrics.py
```diff
@@ -0,0 +1,147 @@
...
+        Instrumentator(
+            should_group_status_codes=True,
+            should_ignore_untemplated=True,
+            excluded_handlers=[".*/_mgmt/.*"],
+        )
+        .add(record_stac_metrics)
+        .instrument(app)
+        .expose(app, endpoint=endpoint, include_in_schema=False)
+    )
+    app.state.stac_metrics_instrumented = True
```
gadomski: Are there cases when we might call `instrument_app` twice? It feels a little strange to modify the app's state just for setting up some middleware?

stac_fastapi/api/stac_fastapi/api/metrics.py
```diff
@@ -0,0 +1,147 @@
...
+    if getattr(app.state, "stac_metrics_instrumented", False):
+        return
+
+    if app.middleware_stack is not None:
+        raise RuntimeError(
+            "Cannot instrument metrics after the application has started. "
+            "Call instrument_app during app construction (before any requests), "
+            "not after startup."
+        )
+
```
gadomski: Doesn't `FastApi` take care of erroring for us (e.g. the CI failure we saw on **stac-fastapi-pgstac**)?

stac_fastapi/api/stac_fastapi/api/metrics.py
```diff
@@ -0,0 +1,147 @@
...
+    return "unknown"
+
+
+def _metric_or_none(factory):
+    try:
+        return factory()
+    except ValueError as exc:
+        if "Duplicated timeseries in CollectorRegistry" not in str(exc):
+            raise
+        return None
```
gadomski: This `try/except` feels a bit strange, what's this for?

gadomski (changes requested):
This feels pretty good to me 👍🏼. I've got some implementation questions/changes, and I'd love to get a review from someone who's using **stac-fastapi** more directly (I'm not really developing against it for any active projects ATM) so requesting from @vincentsarago.

stac_fastapi/api/pyproject.toml
pantierra: Agreed. Dropped the upper bound to >=8.

pyproject.toml
pantierra: Sure, done.

stac_fastapi/api/stac_fastapi/api/app.py
pantierra: Changed metrics to be opt-in with StacApi(add_metrics=True, ...). If enabled and the [metrics] extra isn’t installed, the import fails instead of silently skipping.

stac_fastapi/api/stac_fastapi/api/metrics.py
pantierra: You are probably right. No real need for a double-call guard with construction-time opt-in. Removed the state flag.

stac_fastapi/api/stac_fastapi/api/metrics.py
pantierra: ok, removed the custom check and its test.

stac_fastapi/api/stac_fastapi/api/metrics.py
pantierra: It was only there to swallow duplicate Prometheus registry errors on re-import/re-instrument that I ran into a few times. Not sure if this is actually necessary, removed it.

stac_fastapi/api/stac_fastapi/api/metrics.py
```diff
@@ -0,0 +1,147 @@
+"""Prometheus metrics with low-cardinality STAC operation labels."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from prometheus_client import REGISTRY, Counter, Histogram
+from prometheus_fastapi_instrumentator import Instrumentator
+from prometheus_fastapi_instrumentator.metrics import Info
```
vincentsarago: I would put all the imports in a try/except block 

see https://github.com/developmentseed/titiler/blob/b207d1d6251a046957ee91a7b7cd7579284a9bda/src/titiler/core/titiler/core/telemetry.py#L13-L23

stac_fastapi/api/stac_fastapi/api/app.py
vincentsarago: agreed, I usually do something like 

```python
try:
    import prometheus_fastapi_instrumentator
except ImportError:
    prometheus_fastapi_instrumentator = None # type: ignore

if self.add_metrics:
    assert prometheus_fastapi_instrumentator, "`prometheus_fastapi_instrumentator` needs to be installed to allow metrics"

    from stac_fastapi.api.metrics import instrument_app
   
    instrument_app(self.app)

```

stac_fastapi/api/stac_fastapi/api/metrics.py
pantierra: Done.

jonhealy1:
@vincentsarago @gadomski @pantierra Should this be turned into a proper stac-fastapi extension? It seems like the right way to manage this maybe?

gadomski:
> Should this be turned into a proper stac-fastapi extension? It seems like the right way to manage this maybe?

I _think_ most (all) of the current extensions map onto STAC API extensions, so I'm not sure that metrics would be a fit. There's not really a software reason to not do this, more just a conceptual one.

stac_fastapi/api/stac_fastapi/api/metrics.py
```diff
@@ -0,0 +1,139 @@
...
+        ImportError: If ``stac-fastapi-api[metrics]`` is not installed.
+    """
+    if Instrumentator is None:
+        raise ImportError(
+            "Prometheus metrics require the optional dependency "
+            "`stac-fastapi-api[metrics]`. Install with: "
+            "pip install 'stac-fastapi-api[metrics]'"
+        )
+
+    endpoint = endpoint or metrics_endpoint(app)
```
vincentsarago: I think it will simplify the code if we make `endpoint` required instead of optional (so we can get rid of the metrics_endpoint function).

in app.py#488 we can then set `instrument_app(self.app, endpoint=f"{router_prefix}/_mgmt/metrics")`

stac_fastapi/api/stac_fastapi/api/metrics.py
```diff
@@ -0,0 +1,139 @@
...
+    (
+        Instrumentator(
+            should_group_status_codes=True,
+            should_ignore_untemplated=True,
+            excluded_handlers=[".*/_mgmt/.*"],
+        )
+        .add(record_stac_metrics)
+        .instrument(app)
+        .expose(app, endpoint=endpoint, include_in_schema=False)
+    )
```
vincentsarago, line 131: @pantierra in other PR (titiler-pgstac/tipg) you're registering the app into a frozen set to make sure we don't register the app twice. We don't do it here. 

Q: do we need it / why are we not doing it here?

pantierra:
Just my two-cents: I agree metrics are less a STAC API extension, more like the health-check page: useful ops tooling, and not a STAC feature.

stac_fastapi/api/stac_fastapi/api/metrics.py
pantierra: Yes. Now the code guesses the metrics URL by itself. It’s cleaner if the caller always passes the URL. We can delete the small helper that builds it.

stac_fastapi/api/stac_fastapi/api/metrics.py
pantierra: In `tipg`, metrics are turned on by calling a helper on an already-built app. That helper can get run more than once (tests, factory code, accidental double setup), so tipg keeps a small “already instrumented” list and makes a second call a no-op.

In `stac-fastapi`, metrics would be a construction-time switch (`StacApi(..., add_metrics=True)`), so the library calls the helper once while building the app. Under that path a double-call guard isn’t needed; we skipped it for simplicity after @gadomski noted it earlier.

vincentsarago (approved):

gadomski (approved):

PR #966, opened 2026-08-11
feat: improve security by not returning uncaught exceptions to end-users

StijnCaerts:
**Description:**
By having an exception handler for the Python base `Exception`, all uncaught exception messages will be returned to the end-user. For security purposes, it is better to show a generic error message for uncaught exceptions as these might contain sensitive information. On the other hand, error handling for validating and processing user input can be improved in the application itself, eg. by explicitly raising an `HTTPException` with code `400` when parsing of a filter string fails.

This PR proposes to remove the exception handler for the `Exception` type, as a result end-users will get a generic `HTTP 500` response without detail of the underlying exception raised.

**PR Checklist:**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


jonhealy1 (approved):
LGTM

jonhealy1:
Hi @vincentsarago. What do you think about merging this pr?

